PR #123, opened 2021-03-08
Add Config Configuration Recorder table. Closes #78

### LINKED ISSUE #78: Add table_aws_config_configuration_recorder
Hi team
Would like to see AWS Config table support.

khushboo9024:


aws/table_aws_config_configuration_recorder.go
```diff
@@ -0,0 +1,186 @@
...
+				Hydrate:     getConfigConfigurationRecorderStatus,
+				Transform:   transform.FromValue(),
+			},
+			{
+				Name:        "status_recording",
+				Description: "Specifies whether or not the recorder is currently recording.",
+				Type:        proto.ColumnType_BOOL,
+				Hydrate:     getConfigConfigurationRecorderStatus,
+				Transform:   transform.FromField("Recording"),
+			},
```
LalitTurbot, line 58:
```suggestion
			{
				Name:        "status_recording",
				Description: "Specifies whether or not the recorder is currently recording.",
				Type:        proto.ColumnType_BOOL,
				Hydrate:     getConfigConfigurationRecorderStatus,
				Transform:   transform.FromField("Recording"),
			},
			{
				Name:        "status",
				Description: "The current status of the configuration recorder.",
				Type:        proto.ColumnType_JSON,
				Hydrate:     getConfigConfigurationRecorderStatus,
				Transform:   transform.FromValue(),
			},
```

LalitTurbot: We should try to keep all the JSON fields at end of the table, just before steampipe standard columns

aws-test/tests/aws_config_configuration_recorder/test-list-call-query.sql
```diff
@@ -0,0 +1,2 @@
+select name, role_arn, status, title
+from aws.aws_config_configuration_recorder;
```
LalitTurbot, line 2:
```suggestion
from aws.aws_config_configuration_recorder
where akas::text = '["{{ output.resource_aka.value }}"]'
```
We should have a condition to filter out the results for list calls too

LalitTurbot (changes requested):
@khushboo9024 
few small changes required, otherwise good

cbruno10 (approved):

PR #156, opened 2021-03-11
Update common columns to be returned by global functions, instead of using global variables. Closes #155

### LINKED ISSUE #155: Common columns are shared between tables - this causes issues when joining between 2 tables
As part of the query execution, the sdk updates the column object. Because the common columns are shared global variables, when more than one query is in progress, the data is overwritten, leading to errors such as 
```
Error: column 'region' requires hydrate data from listRoute53Records but none is available
```

The solution is to replace the global variables with global functions which return a copy of the column data

kaidaguerre:
Common columns are shared between tables - this causes issues when joining between 2 tables 

As part of the query execution, the sdk updates the column object. Because the common columns are shared global variables, when more than one query is in progress, the data is overwritten, leading to errors such as
```
Error: column 'region' requires hydrate data from listRoute53Records but none is available
```
The solution is to replace the global variables with global functions which return a copy of the column data

### DIFF
--- a/aws/common_columns.go
+++ b/aws/common_columns.go
@@ -11,79 +11,85 @@ import (
 )
 
 // column definitions for the common columns
-var commonAwsRegionalColumns = []*plugin.Column{
-	{
-		Name:        "partition",
-		Type:        proto.ColumnType_STRING,
-		Hydrate:     getCommonColumns,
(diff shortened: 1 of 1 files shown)


LalitTurbot (approved):
LGTM
I have tested it works fine

PR #284, opened 2021-04-02
fix typo - bad date in CHANGELOG.md

jackdelab:
I think there are some typos in the dates in the changelog, but feel free to deny this PR if I'm wrong.


e-gineer (approved):
Good catch - 20201 would be a long time to wait for a future dated patch!

PR #687, opened 2021-10-14
Add max_concurrent and max_per_second config options

RyanJarv:
I still need to retest this, update the docs, as well as figure out how to run the integration tests. Going ahead and opening this as a work in progress PR for the moment though since it's pretty close to being ready to merge and I'd be interested in suggestions anyone has on this. It's worth noting that this is not a replacement for the backoff request limiting that was merged recently, that change should benefit anyone who uses Steampipe (less likely to crash and plays fairer with the rate-limiting quotas), where this change is targeted twards users that are particularly paranoid about causing disruptions in an their environment.

This adds the max_concurrent and max_per_second config options, both which affect client side rate-limiting of AWS API requests. These settings are useful in cases where you need to ensure the server side rate-limiting won't be triggered. Since AWS shares the API quotas across all services in an account a query against the environment from one user can potentially cause RequestLimitExceeded error responses to be returned from the API to an unrelated service.

In general applications should handle this error code gracefully and in an ideal world this would not cause any problems. However it's not uncommon to come across production apps that will crash when this happens. While you likely want to fix this issue in your code, it is still useful to be able to limit requests when you want to be extra careful in a production environment that you are not familiar with.

The max_concurrent setting here will prevent at most X number of requests from being in flight at the same time. Limiting concurrent requests should prevent any bursty request patterns, going through the allotted requests per second all at the same time. 

Setting max_per_second will allow X number of requests to be made every second. Since AWS does not offer any metrics on the current API quota, it is difficult to know what this needs to be set to, you can however monitor CloudTrail for failed requests and adjust this as needed.

# Integration test logs
TODO: run these
<details>
  <summary>Logs</summary>
  
```
```
</details>

# Example query results
N/A

### DIFF
--- a/aws/connection_config.go
+++ b/aws/connection_config.go
@@ -11,6 +11,8 @@ type awsConfig struct {
 	AccessKey    *string  `cty:"access_key"`
 	SecretKey    *string  `cty:"secret_key"`
 	SessionToken *string  `cty:"session_token"`
+	MaxConcurrent *int      `cty:"max_concurrent"`
+	MaxPerSecond  *int      `cty:"max_per_second"`
 }
 
 var ConfigSchema = map[string]*schema.Attribute{
@@ -29,6 +31,12 @@ var ConfigSchema = map[string]*schema.Attribute{
 	},
 	"session_token": {
 		Type: schema.TypeString,
+	},
+	"max_concurrent": {
+		Type: schema.TypeInt,
+	},
+	"max_per_second": {
+		Type: schema.TypeInt,
 	},
 }
 
--- a/aws/service.go
+++ b/aws/service.go
@@ -84,6 +84,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/wafv2"
 	"github.com/aws/aws-sdk-go/service/wellarchitected"
 	"github.com/aws/aws-sdk-go/service/workspaces"
+	"go.uber.org/ratelimit"
 
 	"github.com/turbot/go-kit/helpers"
 	"github.com/turbot/steampipe-plugin-sdk/plugin"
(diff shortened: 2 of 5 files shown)


cbruno10:
Hey @RyanJarv , sorry for the radio silence on this PR - it's still on my radar to review and add comments, but I haven't gotten a chance yet. I plan to take a look later this week or next, and I'll let you know if I have any questions.

kapilt:
aws api rate limits are per service account region triple, this only operates on a tuple of account, region.

RyanJarv:
To be honest I haven't run into a case where this was needed in a while. I'm sure it will come up again eventually, at which point I might come back to this. For now, considering this is a somewhat uncommon concern, and the fact that retry logic was greatly improved a bit back, I think it's ok to close it for now.

If this does get added I think I would want to remove the extra dependency first, that seems like a bit much for just this.

@kapilt 
> aws api rate limits are per service account region triple, this only operates on a tuple of account, region.
I believe you're right, however I don't believe that's necessary to take into account for a change like this if it adds much complexity (which I suspect is the case). The reasoning being that if you care about this feature it's because you want to avoid impacting other applications in the same account unexpectedly. This concern however can never fully be addressed as there is no way of determining the state of the rate limiting token bucket of the API. In the end it's all relative and a best attempt at playing *extra* nice with other applications so to speak.

cbruno10:
Hey @RyanJarv , thanks again for opening the PR - we like the approach and idea behind providing users an easy way to tune the number of concurrent API requests. We believe that this type of feature would be helpful not only in the AWS plugin, but the others as well, as those services often have strict rate limits.

I've opened https://github.com/turbot/steampipe-plugin-sdk/issues/313 in our SDK repo in an effort to add connection options that can be used with all of the plugins. If you have any additional feedback/suggestions, please feel free to comment in that issue as well.

PR #838, opened 2021-12-17
Add profile from aws spc

dbmurphy:
Extends aws_account to include account_profile column which is pulled get GetConfig from the spc file. This can be useful if someone wants to  join on account and display account_profile vs account_aliases if they don't have aliases defined in AWS. 

# Example query results
<details>
  <summary>Results</summary>
  
```
> select account_id,account_profile from aws_all.aws_account
+---------------+-------------------+
| account_id    | account_profile   |
+---------------+-------------------+
| 000000000000  | shared-services   |
| 000000000001  | nonprod           |
| 000000000002  | prod              |
+---------------+-------------------+

```
</details>

PR #846, opened 2021-12-30
Add profile from aws spc

dbmurphy:
Extends aws_account to include account_profile column which is pulled get GetConfig from the spc file. This can be useful if someone wants to join on account and display account_profile vs account_aliases if they don't have aliases defined in AWS.

Example query results
```
Results
> select account_id,account_profile from aws_all.aws_account
+---------------+-------------------+
| account_id    | account_profile   |
+---------------+-------------------+
| 000000000000  | shared-services   |
| 000000000001  | nonprod           |
| 000000000002  | prod              |
+---------------+-------------------+
````

Updated the code so that during awsConfig setup, if awsConfig.Profile is nil , sets an empty string as the pointer source
```
> select  account_profile,account_id from aws_key.aws_account
+-----------------+--------------+
| account_profile | account_id   |
+-----------------+--------------+
| default         | XXXXXXXXXXXX |
+-----------------+--------------+
Using:

connection "aws_key" {
  plugin      = "aws"
  access_key     = "XXXXXXXXXXXXXXXXXXXX"
  secret_key     = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
  regions     = ["eu-west-1"]
}
```

Additional mixed-mode testing by mixing SSO and Key modes in an aggregator: (SSO profile name and account_ids replaced for account safety.

```
> select account_profile,account_id from aws_test_account_key.aws_account
+------------------------+--------------+
| account_profile        | account_id   |
+------------------------+--------------+
| David Personal Profile | XXXXXXXXXXXX |
+------------------------+--------------+
> select account_profile,account_id from aws_test_db_prod.aws_account
+-----------------------+--------------+
| account_profile       | account_id   |
+-----------------------+--------------+
| David SSO Profile | YYYYYYYYYYYY |
+-----------------------+--------------+
> select account_profile,account_id from aws_test_all.aws_account
+------------------------+--------------+
| account_profile        | account_id   |
+------------------------+--------------+
| David Personal Profile | XXXXXXXXXXXX |
| David SSO Profile  | YYYYYYYYYYYY |
+------------------------+--------------+
> 
```


dbmurphy:
Follow up on #838  which was closed due to corrupting the history when I tried to change my commits emails. 
 @cbruno10 and @rajlearner17   can you re-review

rajlearner17:
@dbmurphy Sure we will be reviewing this and taking the required steps this week. Thank you!

cbruno10:
Hey @dbmurphy , thanks for the PR!

Making this information available definitely makes sense, as it's helpful to see some of the connection information, especially when the rows contain hard to remember information, e.g., AWS account IDs, or when working in a multi-connection environment. We've actually been thinking of how we can make general connection information more available in not only the AWS plugin, but others as well, as there's often a wealth of useful information defined in each connection type.

@johnsmyth Any thoughts on adding a column called `conn_config` (or something similar) that contains the connection config info as a JSON column to all plugin tables as a universal column? Or we can add a new Steampipe table that contains connection information, that queries can then join with in order to pull out useful information?

johnsmyth:
@cbruno10 I agree that it should be more generally applicable.  It would make sense to me to add the connection details in a separate table, either from a plugin or introspection tables, and then add the source connection name column to each table (so you can see the source connection when using an aggregator)

cbruno10:
Thanks for the feedback @johnsmyth , having a universal column that can be tied back to a table seems like a good approach.

@kaidaguerre Any thoughts on the approach mentioned above? Did you have any existing ideas around making connection information available?

dbmurphy:
@johnsmyth @cbruno10 
I think this is more complicated and is part of a conversation about the SSO topic. The current setup provides little ability to really get to that connection's true configuration. We would need to add a whole parsing level to env vars and ~/.aws/* files. 

I say this because exposing the awsConfig struct as a JSON to PG is pretty trivial, or even parsing it into a table of columns. However things get tricky when you may have a key as an ENV input, SPC input, or the SDK reads it directly from the .aws folder in a user home directory.

My goal here was not to go that deep as its a much bigger chunk of work, my goal here was just to expose a "profile" field from the SPC file into aws_account so someone could join on that versus the alias stored in AWS (if they don't use aliases or have access to them). 

Could we maybe get this improvement for now (or ill have to run a custom plugin for all my users with this patch in it). Then expose the whole config at a later data when it can be the finalized config having merged all input types the SDK would have made use of?

johnsmyth:
I propose that we:
  1. Expose the steampipe *connection name* of the source connection in each table.  Ideally, we could add support in the SDK to simplify this across all tables in all plugins.  There may be other steampipe metadata that we would want to expose in the future, so I would propose we add a jsonb column called `sp_ctx` to expose steampipe-related context information, including the source connection name (not the aggregator, but the specific connection that was used to lookup that row). 
  2. Add a new table that exposes the connection details from the steampipe config files.  This could potentially be added to the existing steampipe plugin, or added as a new plugin

  #1 alone would make it easy for the most common case (show me which connection this row came from), as the connection names are likely meaningful to the user.  If you do want to get the profile info from the config you could join the the connection info (from #2) with the connection name (from #1).

@dbmurphy would this meet your needs?  Does this seem reasonable @cbruno10 @kaidaguerre @e-gineer

johnsmyth:
After speaking with @kaidaguerre the connection detail table (#2 above) could also be generated by the SDK, since it already know the connection schema

dbmurphy:
Sorry didnt see this update. @johnsmyth  yes in my case the profile name is enough as in my case our profile names, map to a DNS subdomain for that account in our master account. 

Aka

master = profile foo.com
db_prod =  profile db-prod.foo.com
db_prod =  profile db-non-prod.foo.com
app1_prod =  profile app1.prod.foo.com

So in our case where account alias in AWS is not set up either the SPC config's name or the .aws/profile name both match as are useful. 

The question is should I cancel this PR, and you will do the other as I am not sure where I would get "Expose the steampipe connection name of the source connection in each table." If you provide the guidance I can update this PR to do that not the actual AWS the SDK used, until you can do something more with the SDK itself. 


Given your talking about the SDK, I did share in the other PR#847 [here](https://github.com/turbot/steampipe-plugin-aws/pull/847#issuecomment-1011511650)  My thoughts on the SSO auto login and how I could make it work fairly well except for one case I was stuck on.  Maybe you or @kaidaguerre  can let me know if you agree with my take there, or have a way I missed as you know the SDK better.

e-gineer:
Hey @dbmurphy ... FYI that @kaidaguerre has been working on an SDK update that will automatically add a new context column to every plugin which to include connection information. Feedback welcome!

https://github.com/turbot/steampipe-plugin-sdk/issues/246#issuecomment-1022253636

dbmurphy:
Closing this in favor of @kaidaguerre's work but  the other auto sso stuff needs a comment I have been waiting a month for guidance 👎

PR #914, opened 2022-02-22
Add support for custom retry count closes #913

### LINKED ISSUE #913: Add support for custom retry count
Add support for user can provide the max retry value.

**Additional context**

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>
  
```
N/A
```
</details>

# Example query results
<details>
  <summary>Results</summary>
  
```
N/A
```
</details>

### DIFF
--- a/aws/connection_config.go
+++ b/aws/connection_config.go
@@ -6,13 +6,13 @@ import (
 )
 
 type awsConfig struct {
-	Regions            []string `cty:"regions"`
-	Profile            *string  `cty:"profile"`
-	AccessKey          *string  `cty:"access_key"`
-	SecretKey          *string  `cty:"secret_key"`
-	SessionToken       *string  `cty:"session_token"`
-	MaxErrorRetry      *int     `cty:"max_error_retry"`
-	MinErrorRetryDelay *int     `cty:"min_error_retry_delay"`
+	Regions               []string `cty:"regions"`
+	Profile               *string  `cty:"profile"`
+	AccessKey             *string  `cty:"access_key"`
+	SecretKey             *string  `cty:"secret_key"`
+	SessionToken          *string  `cty:"session_token"`
+	MaxErrorRetryAttempts *int     `cty:"max_error_retry_attempts"`
+	MinErrorRetryDelay    *int     `cty:"min_error_retry_delay"`
 }
 
 var ConfigSchema = map[string]*schema.Attribute{
@@ -32,7 +32,7 @@ var ConfigSchema = map[string]*schema.Attribute{
 	"session_token": {
 		Type: schema.TypeString,
 	},
-	"max_error_retry": {
+	"max_error_retry_attempts": {
 		Type: schema.TypeInt,
 	},
 	"min_error_retry_delay": {
--- a/aws/connection_config.go
+++ b/aws/connection_config.go
@@ -6,12 +6,13 @@ import (
 )
 
 type awsConfig struct {
-	Regions       []string `cty:"regions"`
-	Profile       *string  `cty:"profile"`
-	AccessKey     *string  `cty:"access_key"`
-	SecretKey     *string  `cty:"secret_key"`
-	SessionToken  *string  `cty:"session_token"`
-	MaxRetryCount *string  `cty:"max_retry_count"`
+	Regions            []string `cty:"regions"`
+	Profile            *string  `cty:"profile"`
+	AccessKey          *string  `cty:"access_key"`
+	SecretKey          *string  `cty:"secret_key"`
+	SessionToken       *string  `cty:"session_token"`
+	MaxErrorRetry      *int     `cty:"max_error_retry"`
+	MinErrorRetryDelay *int     `cty:"min_error_retry_delay"`
 }
 
 var ConfigSchema = map[string]*schema.Attribute{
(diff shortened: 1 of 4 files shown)


aws/service.go
```diff
@@ -1763,6 +1764,15 @@ func getSessionWithMaxRetries(ctx context.Context, d *plugin.QueryData, region s
 	// get aws config info
 	awsConfig := GetConfig(d.Connection)
 
+	// Configure the maximum number number of retry user wants to.
+	if awsConfig.MaxRetryCount != nil && *awsConfig.MaxRetryCount != "" {
+		maxRetryValue, err := strconv.Atoi(*awsConfig.MaxRetryCount)
+		if err != nil{
+			panic("Provide an Integer value in spc file for retry count")
```
cbruno10, line 1771: This doesn't seem to be the pattern we use when other config args are incorrect/not set (for example, see below). Where did you find this pattern?

aws/connection_config.go
```diff
@@ -6,11 +6,12 @@ import (
...
-	Profile      *string  `cty:"profile"`
-	AccessKey    *string  `cty:"access_key"`
-	SecretKey    *string  `cty:"secret_key"`
-	SessionToken *string  `cty:"session_token"`
+	Regions       []string `cty:"regions"`
+	Profile       *string  `cty:"profile"`
+	AccessKey     *string  `cty:"access_key"`
+	SecretKey     *string  `cty:"secret_key"`
+	SessionToken  *string  `cty:"session_token"`
+	MaxRetryCount *string  `cty:"max_retry_count"`
```
cbruno10, line 14: Various AWS SDKs use the name `MaxErrorRetry`, so I think we can use the same here, as it's also more descriptive as to what this retry count affects

aws/service.go
```diff
@@ -1763,6 +1764,15 @@ func getSessionWithMaxRetries(ctx context.Context, d *plugin.QueryData, region s
 	// get aws config info
 	awsConfig := GetConfig(d.Connection)
 
+	// Configure the maximum number number of retry user wants to.
+	if awsConfig.MaxRetryCount != nil && *awsConfig.MaxRetryCount != "" {
```
cbruno10, line 1768: I think we should check this in the `getSession` method instead, as we pass in `9` there hardcoded, but it feels like it's better to keep the config check/defaulting to 9 all in one place.

aws/service.go
```diff
@@ -1763,6 +1764,15 @@ func getSessionWithMaxRetries(ctx context.Context, d *plugin.QueryData, region s
...
+		maxRetryValue, err := strconv.Atoi(*awsConfig.MaxRetryCount)
+		if err != nil{
+			panic("Provide an Integer value in spc file for retry count")
+		}
+		maxRetries = maxRetryValue
+	}
```
cbruno10, line 1774: Please check if it makes sense to add a `MinErrorRetryDelay` config arg as well, and if so, please add it.

cbruno10 (changes requested):
@ParthaI Please see comments, thanks!

cbruno10:
@ParthaI Also what doc updates would we need so people know that they can set these, what changing them affects, and what our recommended values are (if any)?

aws/connection_config.go
```diff
@@ -30,6 +32,12 @@ var ConfigSchema = map[string]*schema.Attribute{
 	"session_token": {
 		Type: schema.TypeString,
 	},
+	"max_error_retry": {
```
cbruno10, line 35: Following AWS' environment variable naming `AWS_MAX_ATTEMPTS`, I think let's change to `max_error_retry_attempts` to match more closely with that and to be more descriptive

aws/service.go
```diff
@@ -1749,10 +1749,32 @@ func WorkspacesService(ctx context.Context, d *plugin.QueryData) (*workspaces.Wo
...
-	return getSessionWithMaxRetries(ctx, d, region, 9)
+	awsConfig := GetConfig(d.Connection)
+
+	// As per the logic used in retryRules of NewConnectionErrRetryer with minimum delay of 25ms and maximum
+	// number of retries as 9 (our default), the default maximum delay will not be more than approximately 3 minutes to avoid steampipe
+	// waiting too long to render result
+	maxRetries := 9
+	var minErrRetryDelay time.Duration = 25 * time.Millisecond // Default minimum delay
+
+	// Set max retry count from configuration file
```
cbruno10, line 1760: Can you please add support for the `AWS_MAX_ATTEMPTS` env variable (as detailed in https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-configure) here? The AWS Go SDK v1 doesn't support it, so we should check it here using the follow order of precedence:
- Check config arg, if set use that; else
- Check env variable, if set use that; else
- Fall back to default (9 attempts)

aws/service.go
```diff
@@ -1940,5 +1958,11 @@ func (d ConnectionErrRetryer) RetryRules(r *request.Request) time.Duration {
 	// Creates a new exponential backoff using the starting value of
 	// minDelay and (minDelay * 3^retrycount) * jitter on each failure
 	// as example (23.25ms, 63ms, 238.5ms, 607.4ms, 2s, 5.22s, 20.31s...) up to max.
-	return time.Duration(int(float64(int(minDelay.Nanoseconds())*int(math.Pow(3, float64(retryCount)))) * jitter))
+	// Maximum delay should not be more than 5 min
+	maxDelayTime := time.Duration(int(float64(int(minDelay.Nanoseconds())*int(math.Pow(3, float64(retryCount)))) * jitter))
+	if maxDelayTime > time.Duration(5*time.Minute) {
```
cbruno10, line 1963: Did you do any testing with 3 minutes (our previous max time before)? Is there any difference in performance, or do the subsequent API calls that we retry also succeed when the max is 3 minutes?

cbruno10 (changes requested):
@ParthaI Please see comments, thanks!

cbruno10 (approved):

PR #991, opened 2022-05-12
Add column architectures to aws_lambda_function table Closes #987

### LINKED ISSUE #987: Add column architecture to aws_lambda_function table
Required for AWS Thrifty
[Reference](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.42.25/service/lambda#FunctionConfiguration)

karanpopat:
# Integration test logs
<details>
  <summary>Logs</summary>
  
```
```
</details>

# Example query results
<details>
  <summary>Results</summary>
  
```
> select 
  name,
  runtime, 
  handler, 
  architectures 
from
  aws_lambda_function limit 5;
+---------------------------------------+------------+-------------------------+---------------+
| name                                  | runtime    | handler                 | architectures |
+---------------------------------------+------------+-------------------------+---------------+
| test1                                 | nodejs14.x | index.handler           | ["x86_64"]    |
| tets                                  | nodejs14.x | index.handler           | <null>        |
| cloud9-AppTwo-funcone-W408ZRFMCP4Q    | nodejs6.10 | funcone/index.handler   | ["x86_64"]    |
| cloud9-TestApp-lambdaone-A71A4OKVLX6D | nodejs6.10 | lambdaone/index.handler | ["x86_64"]    |
| graviton2                             | nodejs14.x | index.handler           | ["arm64"]     |
+---------------------------------------+------------+-------------------------+---------------+
```
</details>

### DIFF
--- a/docs/tables/aws_lambda_function.md
+++ b/docs/tables/aws_lambda_function.md
@@ -104,3 +104,16 @@ from
 where
   dead_letter_config_target_arn is null;
 ```
+
+
+### List runtime settings for each lambda function
+
+```sql
+select 
+  name,
+  runtime, 
+  handler, 
+  architectures 
+from
+  aws_lambda_function;
+```
--- a/aws/table_aws_lambda_function.go
+++ b/aws/table_aws_lambda_function.go
@@ -177,6 +177,11 @@ func tableAwsLambdaFunction(_ context.Context) *plugin.Table {
 				Transform:   transform.FromField("Configuration.VpcConfig.VpcId", "VpcConfig.VpcId"),
 			},
 			{
+				Name:        "architectures",
+				Description: "The instruction set architecture that the function supports. Architecture is a string array with one of the valid values.",
+				Type:        proto.ColumnType_JSON,
+			},
+			{
 				Name:        "environment_variables",
 				Description: "The environment variables that are accessible from function code during execution.",
 				Type:        proto.ColumnType_JSON,


docs/tables/aws_lambda_function.md
```diff
@@ -104,3 +104,16 @@ from
 where
   dead_letter_config_target_arn is null;
 ```
+
+
+### List runtime settings for each lambda function
```
cbruno10, line 109:
```suggestion
### List runtime settings for each function
```

We should not include the service name in the description unless we need to differentiate between services

cbruno10 (changes requested):
Please see comment, thanks!

docs/tables/aws_lambda_function.md
```diff
@@ -104,3 +104,16 @@ from
 where
   dead_letter_config_target_arn is null;
 ```
+
```
ParthaI, line 107:
```suggestion
```

karanpopat: I have intentionally kept two line breaks because all the example queries in this doc have two lines

cbruno10: @karanpopat I think early we had 2 line breaks in between examples, but in all newer plugins we're now using only 1. We don't need to go back and update all AWS table docs, but moving forward we should just use 1 and if we're touching a table doc that has 2 line breaks in between, we can adjust those as well

karanpopat: @cbruno10 Got it. Removing the line breaks here then. Thanks!

cbruno10 (approved):

PR #992, opened 2022-05-13
Add `ignore_error_codes` argument to connection config to ignore set of custom AWS error codes

LalitTurbot:


aws/errors.go
```diff
@@ -0,0 +1,60 @@
...
+func isNotFoundError(notFoundErrors []string) plugin.ErrorPredicateWithContext {
+	return func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData, err error) bool {
+		awsConfig := GetConfig(d.Connection)
+
+		// If the get or list hydrate functions have a overriding IgnoreConfig
+		// defined using isNotFoundError function then
+		// It should also check for errors in "ignored_error_codes"
+		allErrors := append(notFoundErrors, awsConfig.IgnoredErrorCodes...)
+		if awsErr, ok := err.(awserr.Error); ok {
+			plugin.Logger(ctx).Info("isNotFoundError", "AWS Error CODE", awsErr.Code())
```
cbruno10, line 21: Do we still want to include these as part of the release? Or do we need to modify the error message/change the level?

aws/errors.go
```diff
@@ -0,0 +1,60 @@
...
+
+// shouldIgnoreErrorPluginDefault:: Plugin level default function to ignore a set errors for hydrate functions based on `ignored_error_codes`
+func shouldIgnoreErrorPluginDefault() plugin.ErrorPredicateWithContext {
+	return func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData, err error) bool {
+		if !hasIgnoredErrorCodes(d.Connection) {
+			return false
+		}
+		awsConfig := GetConfig(d.Connection)
+		if awsErr, ok := err.(awserr.Error); ok {
+			plugin.Logger(ctx).Info("shouldIgnoreErrorPluginDefault", "AWS Error CODE", awsErr.Code())
```
cbruno10, line 43: Same comment as above

cbruno10 (changes requested):
@LalitTurbot I didn't do a full review yet, but did a quick scan for log messages/default error codes

cbruno10 (approved):

PR #1082, opened 2022-06-14
Add s3_force_path_style config arg to force s3 path-style addressing…

### LINKED ISSUE #1077: Localstack: Force S3 to use `path` style
### The problem
Local stack requires to force S3 path to `path` style, whereas by default Go AWS SDK uses `auto` that actually means `virtual`. 
That behavior causes failures by querying S3-related information over the Localstack. Details can be found:
- [ S3 path styles](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html)
- [related Localstack issue](https://github.com/localstack/localstack/issues/1180)
- failure details are described in the [How to reproduce](#How-to-reproduce)

### Desired solution
It is required to introduce an ability to set the boolean value to [S3ForcePathStyle  in aws.Config](https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config).
That can be done by adding the correspondent connection parameter `s3_force_path_style`.

### The alternative solution
Potentially there is an alternative solution that I tried and it is actually does not work at all. 

AWS provides an ability to specify [`addressing_style` in aws credentials file]( https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-s3):
```
Specifies which addressing style to use. This controls whether the bucket name is in the hostname or is part of the URL. Valid values are: path, virtual, and auto. The default value is auto.

There are two styles of constructing an Amazon S3 endpoint. The first is called virtual and includes the bucket name as part of the hostname. For example: https://bucketname.s3.amazonaws.com. Alternatively, with the path style, you treat the bucket name as if it is a path in the URI; for example, https://s3.amazonaws.com/bucketname. The default value in the CLI is to use auto, which attempts to use the virtual style where it can, but will fall back to path style when required. For example, if your bucket name is not DNS compatible, the bucket name cannot be part of the hostname and must be in the path. With auto, the CLI will detect this condition and automatically switch to path style for you. If you set the addressing style to path, you must then ensure that the AWS Region you configured in the AWS CLI matches the Region of your bucket.
```

I have created an AWS profile:
```
[test]
s3 = 
  addressing_style = path
```

Used this profile in AWS steampipe connection:
```
connection "aws" {
   plugin="aws"
   access_key="default_access_key"
   endpoint_url="http://localhost:4566"
   regions=["sa-east-1"]
   secret_key="default_secret_key"
   profile="test"
```

And got the same error.

### How to reproduce

1. [Run Localstack](https://docs.localstack.cloud/get-started/)
2. Configure AWS connection for steampipe providing `endpoint_url` parameter.
Here is an example:
```
connection "aws" {
   plugin="aws"
   access_key="default_access_key"
   endpoint_url="http://localhost:4566"
   regions=["sa-east-1"]
   secret_key="default_secret_key"
```
3. Execute the following query:
```
select name, arn, region, creation_date as create_time, 
  jsonb_path_query(server_side_encryption_configuration, '$.Rules[0].ApplyServerSideEncryptionByDefault') ->> 'KMSMasterKeyID' as kms_key_id
from aws_s3_bucket
union
select name, arn, region, creation_date as create_time,
  null as kms_key_id
from aws_s3_bucket
where server_side_encryption_configuration is null
```

The expected result: query fails with:
```
Error: RequestError: send request failed
caused by: Get "http://lab-bucket-1.localhost:4566/?encryption=": dial tcp 120.78.181.57:4566: i/o timeout (SQLSTATE HV000)
```

Plugin logs:
```
  2022-06-13 13:10:07.552 UTC [ERROR] steampipe-plugin-aws.plugin: [ERROR] 1655125176492: table 'aws_s3_bucket' failed to get column data, callId 1655125176492: table 'aws_s3_bucket' column 'server_side_encryption_configuration' requires hydrate data from getBucketEncryption, which failed with error RequestError: send request failed
  2022-06-13 13:10:16.072 UTC [ERROR] steampipe-plugin-aws.plugin: [ERROR] 1655125176492: callHydrate getBucketEncryption finished with error: RequestError: send request failed
  2022-06-13 13:10:16.072 UTC [WARN]  steampipe-plugin-aws.plugin: [WARN]  1655125176492: hydrate error chan select error RequestError: send request failed
  2022-06-13 13:10:16.072 UTC [WARN]  steampipe-plugin-aws.plugin: [WARN]  1655125176492: getRow failed with error RequestError: send request failed
  2022-06-13 13:10:16.072 UTC [ERROR] steampipe-plugin-aws.plugin: [ERROR] 1655125176492: table 'aws_s3_bucket' failed to get column data, callId 1655125176492: table 'aws_s3_bucket' column 'server_side_encryption_configuration' requires hydrate data from getBucketEncryption, which failed with error RequestError: send request failed
```  

**The root cause:** 
`getBucketEncryption` uses improper (from the Localstack perspective) url `http://**lab-bucket-2.**localhost:4566/?encryption=' 
instead of `localhost:4566/lab-bucket-2?encryption=`.

srgg:
… closes #1077

# Integration test logs
<details>
  <summary>Logs</summary>
  
```
```
</details>

# Example query results
<details>
  <summary>Results</summary>
  
```
```
</details>

### DIFF
--- a/config/aws.spc
+++ b/config/aws.spc
@@ -36,7 +36,6 @@ connection "aws" {
 
   # Set to `true` to force S3 requests to use path-style addressing,
   # i.e., `http://s3.amazonaws.com/BUCKET/KEY`. By default, the S3 client
-  # will use virtual hosted bucket addressing when possible
-  # (`http://BUCKET.s3.amazonaws.com/KEY`).
-  #s3_force_path_style = true
+  # will use virtual hosted bucket addressing when possible (`http://BUCKET.s3.amazonaws.com/KEY`).
+  #s3_force_path_style = false
 }
--- a/config/aws.spc
+++ b/config/aws.spc
@@ -33,4 +33,10 @@ connection "aws" {
   # If not set, the default AWS generated endpoint will be used.
   # Can also be set with the AWS_ENDPOINT_URL environment variable.
   #endpoint_url = "http://localhost:4566"
+
+  # Set to `true` to force S3 requests to use path-style addressing,
+  # i.e., `http://s3.amazonaws.com/BUCKET/KEY`. By default, the S3 client
+  # will use virtual hosted bucket addressing when possible
+  # (`http://BUCKET.s3.amazonaws.com/KEY`).
+  #s3_force_path_style = true
 }
--- a/docs/index.md
+++ b/docs/index.md
@@ -105,9 +105,8 @@ connection "aws" {
   
   # Set to `true` to force S3 requests to use path-style addressing,
   # i.e., `http://s3.amazonaws.com/BUCKET/KEY`. By default, the S3 client
-  # will use virtual hosted bucket addressing when possible
-  # (`http://BUCKET.s3.amazonaws.com/KEY`).
-  #s3_force_path_style = true
+  # will use virtual hosted bucket addressing when possible (`http://BUCKET.s3.amazonaws.com/KEY`).
+  #s3_force_path_style = false
 }
 ```
 
@@ -120,7 +119,7 @@ connection "aws" {
 - `regions` - (Optional) List of AWS regions Steampipe will connect to. Can also be set with the `AWS_REGION` or `AWS_DEFAULT_REGION` environment variables, or the region specified in the active profile.
 - `secret_key` - (Optional) AWS secret key. Can also be set with the `AWS_SECRET_ACCESS_KEY` environment variable.
 - `session_token` - (Optional) Session token for validating temporary credentials. Can also be set with the `AWS_SESSION_TOKEN` environment variable.
-- `s3_force_path_style`- (Optional) The only option to force AWS S3 client to use `path-style` addressing. By default, the S3 client will use virtual hosted bucket addressing when possible.
+- `s3_force_path_style`- (Optional) Specifies whether to use path-style addressing, i.e., `https://s3.amazonaws.com/BUCKET/KEY`, or virtual hosted bucket addressing, i.e., `https://BUCKET.s3.amazonaws.com/KEY`. By default, the S3 client will use virtual hosted bucket addressing when possible.
 
 By default, all options are commented out in the default connection, thus Steampipe will resolve your region and credentials using the same mechanism as the AWS CLI (AWS environment variables, default profile, etc).  This provides a quick way to get started with Steampipe, but you will probably want to customize your experience using configuration options for [querying multiple regions](#multi-region-connections), [configuring credentials](#configuring-aws-credentials) from your [AWS Profiles](#aws-profile-credentials), [SSO](#aws-sso-credentials), [aws-vault](#aws-vault-credentials) etc.
 
@@ -371,7 +370,7 @@ connection "aws_account_123456789012" {
 
 ### IAM Access Key Pair Credentials
 
-сThe AWS plugin allows you set static credentials with the `access_key`, `secret_key`, and `session_token` arguments in your connection.
+The AWS plugin allows you set static credentials with the `access_key`, `secret_key`, and `session_token` arguments in your connection.
 
 ```hcl
 connection "aws_account_x" {
--- a/docs/index.md
+++ b/docs/index.md
@@ -102,6 +102,12 @@ connection "aws" {
   # If not set, the default AWS generated endpoint will be used.
   # Can also be set with the AWS_ENDPOINT_URL environment variable.
   #endpoint_url = "http://localhost:4566"
+  
+  # Set to `true` to force S3 requests to use path-style addressing,
+  # i.e., `http://s3.amazonaws.com/BUCKET/KEY`. By default, the S3 client
+  # will use virtual hosted bucket addressing when possible
+  # (`http://BUCKET.s3.amazonaws.com/KEY`).
+  #s3_force_path_style = true
 }
 ```
 
@@ -114,6 +120,7 @@ connection "aws" {
 - `regions` - (Optional) List of AWS regions Steampipe will connect to. Can also be set with the `AWS_REGION` or `AWS_DEFAULT_REGION` environment variables, or the region specified in the active profile.
 - `secret_key` - (Optional) AWS secret key. Can also be set with the `AWS_SECRET_ACCESS_KEY` environment variable.
 - `session_token` - (Optional) Session token for validating temporary credentials. Can also be set with the `AWS_SESSION_TOKEN` environment variable.
+- `s3_force_path_style`- (Optional) The only option to force AWS S3 client to use `path-style` addressing. By default, the S3 client will use virtual hosted bucket addressing when possible.
 
 By default, all options are commented out in the default connection, thus Steampipe will resolve your region and credentials using the same mechanism as the AWS CLI (AWS environment variables, default profile, etc).  This provides a quick way to get started with Steampipe, but you will probably want to customize your experience using configuration options for [querying multiple regions](#multi-region-connections), [configuring credentials](#configuring-aws-credentials) from your [AWS Profiles](#aws-profile-credentials), [SSO](#aws-sso-credentials), [aws-vault](#aws-vault-credentials) etc.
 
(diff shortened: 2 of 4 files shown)


cbruno10:
Thanks @srgg for another PR!

In the issue, I can see that some columns from the `aws_s3_bucket` table return ok (assuming that was run against your LocalStack environment), but some of the other S3 table queries have failed.

Can you also please run `select * from aws_s3_bucket`, as there are many columns that have separate hydrates, so making sure all of these work would be helpful.

For the other S3 tables that have failed, do you know if other users of LocalStack have reported similar issues when using the S3 path style addressing?

Another set of tests that would be worthwhile IMO is to run `select * from aws_s3_...` tables against a live AWS account with `s3_force_path_style = true` and see if all 3 queries run successfully in this case as well, or if the access point + account settings table queries also fail there.
 
 If you have any questions, please let me know, thanks!

srgg:
@cbruno10

> In the issue, I can see that some columns from the `aws_s3_bucket` table return ok (assuming that was run against your LocalStack environment), but some of the other S3 table queries have failed.

That's correct

> Can you also please run `select * from aws_s3_bucket`, as there are many columns that have separate hydrates, so making sure all of these work would be helpful.

 I've missed the point of making '*' query from `aws_s3_bucket`, I checked and found  it works on my localstack

> For the other S3 tables that have failed, do you know if other users of LocalStack have reported similar issues when using the S3 path style addressing?

I did a brief investigation, it seems that the root cause is the AWS GO SDK that does not work properly, I will provide more details after I get evidence.
 
> Another set of tests that would be worthwhile IMO is to run `select * from aws_s3_...` tables against a live AWS account with `s3_force_path_style = true` and see if all 3 queries run successfully in this case as well, or if the access point + account settings table queries also fail there.

Notsute that I can that, but I will try.

srgg:
@cbruno10 

Each of the following queries relies on particular, means separate,  AWS API: 

1. select * from aws_s3_bucket relies on [AWS S3 API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Types_Amazon_Simple_Storage_Service.html)
2.  select * from aws_s3_access_point relies on [AWS S3 Control API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Types_AWS_S3_Control.html)
3. select * from aws_s3_account_settings relies on [AWS S3 on Outposts API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Types_AWS_S3_Control.html)

Therefore, in AWS SDK each of these AWS APIs has its own clients:
  1. [Go client for S3 API](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/)
  4. [Go client for S3 Control API](https://docs.aws.amazon.com/sdk-for-go/api/service/s3control/)
  5. [GO client for S3 on Outpost API](https://docs.aws.amazon.com/sdk-for-go/api/service/s3outposts/)

That is why the proposed change is only applicable to the AWS S3 API and limited by the S3 client and this limitation is dictated by AWS.

Since at the moment there are no known ways to make queries (2) and (3) work with  Localstack, I would like to suggest merging this change. And address the rest of the problems as soon as correspondent request will be created and/or when there will be ability to fix them

What is also worth to mention - Localstack has kind of a community slack chanel, but it is required to be invited, and I do not have one, so I can't bring this issues to the community.

cbruno10:
@srgg Thanks for looking into it! So if I understand correctly, even without any of the changes in this PR, the S3 access point and account settings queries do **not** work against LocalStack environments?

Also, did you get a chance yet to try running `select * from aws_s3_bucket` against a live AWS environment with `s3_force_path_style = true`? Thanks!

e-gineer:
I believe Path style addressing for S3 has been deprecated? Seems like something we should not move to by default.

https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

srgg:
@cbruno10 
> @srgg Thanks for looking into it! So if I understand correctly, even without any of the changes in this PR, the S3 access point and account settings queries do **not** work against LocalStack environments?
> 

Yes, with or without changes last two queries do not work against Localstack.

> Also, did you get a chance yet to try running `select * from aws_s3_bucket` against a live AWS environment with `s3_force_path_style = true`? Thanks!

Yes, I did. It works against live AWS


@e-gineer 
> I believe Path style addressing for S3 has been deprecated? Seems like something we should not move to by default.
> https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

AWS has to deprecate, but [AWS Delays “Path-Style” S3 Deprecation at Last Minute](https://techmonitor.ai/technology/cloud/aws-s3-path-deprecation)


By default it will use AWS default that means do not force path style

cbruno10:
@e-gineer Agree, since by default the CLI and SDKs' `auto` mode is to use virtual style pathing, in this plugin, we'd continue to use the auto/virtual style pathing. Only if a user set `s3_force_path_style = true` would it switch from virtual -> path style. Is this what you meant?

srgg:
@cbruno10 follow up

docs/index.md
```diff
@@ -114,6 +120,7 @@ connection "aws" {
 - `regions` - (Optional) List of AWS regions Steampipe will connect to. Can also be set with the `AWS_REGION` or `AWS_DEFAULT_REGION` environment variables, or the region specified in the active profile.
 - `secret_key` - (Optional) AWS secret key. Can also be set with the `AWS_SECRET_ACCESS_KEY` environment variable.
 - `session_token` - (Optional) Session token for validating temporary credentials. Can also be set with the `AWS_SESSION_TOKEN` environment variable.
+- `s3_force_path_style`- (Optional) The only option to force AWS S3 client to use `path-style` addressing. By default, the S3 client will use virtual hosted bucket addressing when possible.
```
cbruno10, line 123:
```suggestion
- `s3_force_path_style`- (Optional) Specifies whether to use path-style addressing, i.e., `https://s3.amazonaws.com/BUCKET/KEY`, or virtual hosted bucket addressing, i.e., `https://BUCKET.s3.amazonaws.com/KEY`. By default, the S3 client will use virtual hosted bucket addressing when possible.
```

config/aws.spc
```diff
@@ -33,4 +33,10 @@ connection "aws" {
   # If not set, the default AWS generated endpoint will be used.
   # Can also be set with the AWS_ENDPOINT_URL environment variable.
   #endpoint_url = "http://localhost:4566"
+
+  # Set to `true` to force S3 requests to use path-style addressing,
+  # i.e., `http://s3.amazonaws.com/BUCKET/KEY`. By default, the S3 client
+  # will use virtual hosted bucket addressing when possible
+  # (`http://BUCKET.s3.amazonaws.com/KEY`).
```
cbruno10, line 40:
```suggestion
  # will use virtual hosted bucket addressing when possible (`http://BUCKET.s3.amazonaws.com/KEY`).
```

It's a little long, but I think that's more readable than pushing just the example to the next line

docs/index.md
```diff
@@ -102,6 +102,12 @@ connection "aws" {
   # If not set, the default AWS generated endpoint will be used.
   # Can also be set with the AWS_ENDPOINT_URL environment variable.
   #endpoint_url = "http://localhost:4566"
+  
+  # Set to `true` to force S3 requests to use path-style addressing,
```
cbruno10, line 106: Please remember to make any similar changes from `aws.spc` here as well

cbruno10 (changes requested):
Please see suggestions, thanks!

docs/index.md
```diff
@@ -364,7 +371,7 @@ connection "aws_account_123456789012" {
 
 ### IAM Access Key Pair Credentials
 
-The AWS plugin allows you set static credentials with the `access_key`, `secret_key`, and `session_token` arguments in your connection.
+сThe AWS plugin allows you set static credentials with the `access_key`, `secret_key`, and `session_token` arguments in your connection.
```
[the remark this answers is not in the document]
srgg: Done

config/aws.spc
```diff
@@ -33,4 +33,10 @@ connection "aws" {
...
+  #s3_force_path_style = true
```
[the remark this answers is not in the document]
srgg: Done

docs/index.md
srgg: Done

config/aws.spc
srgg: Done

docs/index.md
srgg: Done

cbruno10:
Thanks again @srgg for a great pull request! We'll release this in a new plugin version later this week.

PR #1186, opened 2022-07-22
Update plugin to use `github.com/aws/aws-sdk-go-v2`- Week1

LalitTurbot:
## Other changes for SDK migration
- [x] Tested - all authentication working with version 2 as was with version 1 of the SDK.
- [x] Updated error handling to work with v1 and v2. - tested
- [x] Tested with [AWS Localstack](https://localstack.cloud/)  - worked without any issue
- [x] Retry logic is working fine - tested with the aggregate connection of 10 accounts for tables like aws_iam_policy and aws_iam_role (tables with 1000+ records).

## Tables Migrated
1. aws_iam_access_advisor
   - [x] Old and new data matched
   - [x] Paging tested
   - [x] Limit tested
   - [x] N/A no mod queries using it

2. aws_iam_access_key
   - [x] Old and new data matched
   - [x] Paging tested
   - [x] Limit tested
   - [x] Compliance mods queries using this table tested

3. aws_iam_account_password_policy
   - [x] Old and new data matched
   - [x] Paging - N/A due to not supported
   - [x] Limit - N/A due to single row always returned
   - [x] Compliance mods queries using this table tested

4. aws_iam_account_summary
   - [x] Old and new data matched
   - [x] Paging - N/A due to not supported
   - [x] Limit - N/A due to single row always returned
   - [x] Compliance mods queries using this table tested

5. aws_iam_credential_report
   - [x] Old and new data matched
   - [x] Paging - N/A due to not supported
   - [x] Limit - N/A due to single row always returned
   - [x] Compliance mods queries using this table tested

6. aws_iam_group
   - [x] Old and new data matched
   - [x] Paging tested
   - [x] Limit tested
   - [x] Compliance mods queries using this table tested

7. aws_iam_policy
   - [x] Old and new data matched
   - [x] Paging tested
   - [x] Limit tested
   - [x] Compliance mods queries using this table tested

8. aws_iam_policy_attachment
   - [x] Old and new data matched
   - [x] Paging tested (uses list policies as parent hydrate)
   - [x] Limit tested
   - [x] N/A - no mod queries using it

9. aws_iam_policy_simulator
   - [x] Old and new data matched
   - [x] Paging - N/A due to not supported
   - [x] Limit - N/A due to not being supported
   - [x] N/A - no mod queries using it

10. aws_iam_role
    - [x] Old and new data comparison
    - [x] Paging tested
    - [x] Limit tested
    - [x] Compliance, Perimeter, Tags mods queries using this table tested

11. aws_iam_saml_provider
    - [x] Old and new data comparison
    - [x] Paging is not applicable
    - [x] Limit
    - [x] N/A - no mod queries using it

12. aws_iam_server_certificate
    - [x] Old and new data comparison
    - [x] Paging tested
    - [x] Limit tested
    - [x] Compliance and Tags mods queries using this table tested

13. aws_iam_user
    - [x] Old and new data comparison
    - [x] Paging tested
    - [x] Limit tested
    - [x] Compliance and Tags mods queries using this table tested

14. aws_iam_virtual_mfa_device
    - [x] Old and new data comparison
    - [x] Paging tested
    - [x] Limit tested
    - [x] Compliance mod queries using this table tested

15. aws_s3_bucket
    - [x] Old and new data comparison
    - [x] Paging is not applicable
    - [x] Limit tested
    - [x] Compliance, Perimeter, Tags mods queries using this table tested

16. aws_sns_topic
    - [x] Old and new data comparison
    - [x] Paging tested
    - [x] Limit tested
    - [x] Compliance and Perimeter mods queries using this table tested

17. aws_acm_certificate
    - [x] Old and new data comparison
    - [x] Paging
    - [x] Limit
    - [x] Compliance mods queries using this table tested

18. aws_api_gateway_api_authorizer

    - [x] Old and new data comparison
    - [x] Paging is not applicable
    - [x] Limit
    - [x] N/A - no mod queries using it

19. aws_api_gateway_api_key

    - [x] Old and new data comparison
    - [x] Paging
    - [x] Limit
    - [x] N/A - no mod queries using it

20. aws_api_gateway_rest_api

    - [x] Old and new data comparison
    - [x] Paging
    - [x] Limit
    - [x] Compliance and Perimeter mods queries using this table tested

21. aws_api_gateway_stage

    - [x] Old and new data comparison
    - [x] Paging is not applicable
    - [x] Limit is not applicable
    - [x] Compliance and tags mods queries using this table tested

22. aws_api_gateway_usage_plan

    - [x] Old and new data comparison
    - [x] Paging
    - [x] Limit
    - [x] N/A - no mod queries using it

22. aws_api_gatewayv2_api

    - [x] Old and new data comparison
    - [x] Paging retested
    - [x] Limit retested
    - [x] N/A - no mod queries using it

23. aws_api_gatewayv2_domain_name

    - [x] Old and new data comparison
    - [x] Paging retested
    - [x] Limit retested

24. aws_api_gatewayv2_integration

    - [x] Old and new data comparison
    - [x] Paging retested
    - [x] Limit retested

25. aws_api_gatewayv2_stage

    - [x] Old and new data comparison
    - [x] Paging retested
    - [x] Limit retested

26. aws_dynamodb_backup

    - [x] Old and new data comparison
    - [x] Paging is not applicable
    - [x] Limit
    - [x] N/A - no mod queries using it


aws/table_aws_sns_topic.go
```diff
@@ -288,45 +291,50 @@ func getTopicAttributes(ctx context.Context, d *plugin.QueryData, h *plugin.Hydr
 		arn = d.KeyColumnQuals["topic_arn"].GetStringValue()
 	}
 
-	// Create session
-	svc, err := SNSService(ctx, d)
+	if arn == "" {
```
cbruno10, line 294: What case does this cover?

aws/table_aws_sns_topic.go
```diff
@@ -336,15 +344,23 @@ func listTagsForSnsTopic(ctx context.Context, d *plugin.QueryData, h *plugin.Hyd
...
+
+	turbotTagsMap := map[string]string{}
+	for _, i := range tags.Tags {
+		turbotTagsMap[*i.Key] = *i.Value
 	}
+
 	return turbotTagsMap, nil
 }
+
+func snsTopicSrcTags(_ context.Context, d *transform.TransformData) (interface{}, error) {
```
cbruno10, line 359: Is this what we normally name these types of transform functions?

aws/table_aws_s3_bucket.go
```diff
@@ -145,7 +147,7 @@ func tableAwsS3Bucket(_ context.Context) *plugin.Table {
 				Description: "A container for specifying the notification configuration of the bucket. If this element is empty, notifications are turned off for the bucket.",
 				Type:        proto.ColumnType_JSON,
 				Hydrate:     getS3BucketEventNotificationConfigurations,
-				Transform:   transform.FromValue(),
+				Transform:   transform.FromValue().Transform(removeMedataFromS3BucketEventNotificationConfigurations),
```
cbruno10, line 150:
```suggestion
				Transform:   transform.FromValue().Transform(removeMetadataFromS3BucketEventNotificationConfigurations),
```

cbruno10: Can you please search for any instances of misspelled `metadata` words, e.g., `Medata`?

cbruno10, line 150: Is this something we should be removing through the transforms, or should we only return the data we want (with one of our own structs) in the `getS3BucketEventNotificationConfigurations` function?

aws/table_aws_iam_user.go
```diff
@@ -483,3 +505,21 @@ func userMfaStatus(_ context.Context, d *transform.TransformData) (interface{},
 
 	return false, nil
 }
+
+func userMfaDevices(_ context.Context, d *transform.TransformData) (interface{}, error) {
```
cbruno10, line 509: Is this what we normally call transform functions? I would expect some verb, e.g., `get`, `list`, `format`, in the function name, as I'm not sure what output I'd expect just from the function name.

cbruno10 (changes requested):
@LalitTurbot @ParthaI Using this branch, have all of the queries in the AWS mods that use these tables been run (with actual data in the accounts/regions we run against)? It may be easier to run `steampipe check all`, rather than looking for each query, though it may take a while, especially in the AWS Compliance mod.

And based on the PR body, it didn't look like there were any breaking changes, but as a final sign off, can you please confirm there are no breaking changes to any columns in any of the modified tables?

aws/utils.go
```diff
@@ -73,7 +73,7 @@ func extractNameFromSqsQueueURL(queue string) (string, error) {
 }
 
 func handleNilString(_ context.Context, d *transform.TransformData) (interface{}, error) {
-	value := types.SafeString(d.Value)
+	value := types.SafeString(fmt.Sprintf("%v", d.Value))
```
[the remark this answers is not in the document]
LalitTurbot: The issue in the AWS v2 SDK. AWS changed the data type from `*string` or  `string` to Package Specific Struct
Particularly, in the S3 table to service specific types which indirectly is a string type. This handles that conversion and has also verified wherever we have used `types.SafeString` function

aws/errors.go
```diff
@@ -25,6 +27,16 @@ func isNotFoundError(notFoundErrors []string) plugin.ErrorPredicateWithContext {
 				}
 			}
 		}
+
+		var ae smithy.APIError
```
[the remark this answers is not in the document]
LalitTurbot: Yes, we are using the `errors` package both with v1 and v2 tables. 
Generally, we are using it two convert the API error to the relevant error type.

Also, One observation around now with v2 SDK this format of conversion `err.(smithy.APIError)` doesn't work instead we need to use 
```go
      var ae smithy.APIError
      if errors.As(err, &ae) {
```
https://aws.github.io/aws-sdk-go-v2/docs/handling-errors/#api-error-responses

aws/service_v2.go
```diff
@@ -0,0 +1,350 @@
...
+// BackoffDelay returns the duration to wait before the next attempt should be
+// made. Returns an error if unable get a duration.
+func (j *ExponentialJitterBackoff) BackoffDelay(attempt int, err error) (time.Duration, error) {
+	minDelay := j.minDelay
+
+	// The calculatted jitter will be between [0.8, 1.2)
+	var jitter = float64(rand.Intn(120-80)+80) / 100
+
+	retryTime := time.Duration(int(float64(int(minDelay.Nanoseconds())*int(math.Pow(3, float64(attempt)))) * jitter))
+	// log.Printf("[INFO] *******INSIDE CODE********** Attempt: %d, DelayInSeconds: %f, Delay: %v", attempt, retryTime.Seconds(), retryTime)
```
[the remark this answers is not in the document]
LalitTurbot: We have not logged it earlier too, as it leads to much noise in the logs file. 
I can remove it from here, but I just wanted to keep it commented as it helps in future debugging scenarios.

aws/table_aws_s3_bucket.go
LalitTurbot: I will fix this

LalitTurbot: As a standard, which I remember from my previous conversations with John. It is better not to format the output of Hydrate API calls internally in the Hydrate function, instead use Transforms to achieve it.

aws/table_aws_sns_topic.go
LalitTurbot: Generally, transform functions are the ones which don't have API calls inside them. Instead, they are used to convert and sanitise the outputs.

Signature of transform function is
`func funcname(context.Context, *transform.TransformData) (interface{}, error)`

aws/table_aws_iam_user.go
LalitTurbot: Agree, I will use the verbs like `format`, `convert`, etc, as we usually transform API outputs in these function instead, of getting or listing

aws/utils.go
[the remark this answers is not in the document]
cbruno10: Have we tested this with v1 tables that use this function, or is it only used in the S3 bucket table?

aws/errors.go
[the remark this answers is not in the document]
cbruno10: To avoid potentially changing v1 table error handling, how much effort would it be to create new versions of functions we use in `errors.go` in the `errors_v2.go` function, and then have the updated v2 tables use that function instead?

aws/service_v2.go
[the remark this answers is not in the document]
cbruno10: I think I'm ok with leaving it in, but can you please convert it to a proper logging statement (probably at TRACE level) so if someone does re-enable it while locally testing, then it'll appear in the format other messages also appear?

aws/table_aws_sns_topic.go
```diff
@@ -336,15 +344,23 @@ func listTagsForSnsTopic(ctx context.Context, d *plugin.QueryData, h *plugin.Hyd
...
+
+	turbotTagsMap := map[string]string{}
+	for _, i := range tags.Tags {
+		turbotTagsMap[*i.Key] = *i.Value
 	}
+
 	return turbotTagsMap, nil
 }
+
+func formatSnsTopicEmptyTags(_ context.Context, d *transform.TransformData) (interface{}, error) {
```
cbruno10, line 359: What formatting are we doing in this function? Here the function name starts with `format`, while in `aws_iam_user` table, the transform functions start with `handle`. Should this one also use the word `handle` to be consistent, or is there some formatting occurring?

aws/table_aws_s3_bucket.go
cbruno10: What's the advantage of doing it through a transform function, because the code in the transform function almost looks the same (create a struct/map, create properties it in representing those we want from the value, and then returning it)?

aws/table_aws_s3_bucket.go
```diff
@@ -159,7 +160,7 @@ func tableAwsS3Bucket(_ context.Context) *plugin.Table {
 				Description: "The access control list (ACL) of a bucket.",
 				Type:        proto.ColumnType_JSON,
 				Hydrate:     getBucketACL,
-				Transform:   transform.FromValue(),
+				Transform:   transform.FromValue().Transform(removeMetadataFromBucketAclOutput),
```
cbruno10, line 163:
```suggestion
				Transform:   transform.FromValue().Transform(removeMetadataFromBucketACLOutput),
```

cbruno10 (changes requested):
@LalitTurbot Please see my responses on the previous review comments, along with new ones. Thanks!

aws/errors.go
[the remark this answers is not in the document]
LalitTurbot: `isNotFoundError` and `shouldIgnoreErrorPluginDefault` error handling functions are used at the plugin level too
https://github.com/turbot/steampipe-plugin-aws/blob/update-aws-go-sdk/aws/plugin.go#L23

aws/service_v2.go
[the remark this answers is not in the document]
LalitTurbot: For now, I have removed it as I could not bring the called hydrate function details into it. That would help in debugging later.

aws/utils.go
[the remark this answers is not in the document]
LalitTurbot: Yes

aws/table_aws_s3_bucket.go
LalitTurbot: From my perspective, I don't see any added advantage the handling will be the same even if choose to remove the transform and handle it in the hydrate function. 
The only disadvantage will be anyone who wants to use the raw output of the hydrate function further, he/she couldn't do so as we are already limiting the output in the transform function. Though metadata doesn't seem to be much useful in that case.

aws/table_aws_sns_topic.go
cbruno10: As per discussion, we'll change this function name to be `handleSnsTopicEmptyTags` to be consistent

LalitTurbot: Will go with `handle` as a prefix for the function.

cbruno10 (approved):

PR #1206, opened 2022-08-13
feat: added aws_ses_domain_identity table - resolves #1203

### LINKED ISSUE #1203: Add table aws_ses_domain_identity or aws_ses_identity
Currently the Steampipe AWS plugin has the table `aws_ses_email_identity`, to also query SES domain identities I would like to add a new table.
I'm still not sure which option would be the best, so I'm interested in your feedback.

## Ideas

These two tables would be an idea:

1. aws_ses_domain_identity

Similar to the existing `aws_ses_email_identity` table a new table would be added which filters for `Domain` instead of `EmailAddress`

2. aws_ses_identity

Instead of having two separate tables with pre applied filters, a general SES identity table is added without the `IdentityType` filter, filtering can then be done via SQL in SteamPipe

## References

- API reference: https://docs.aws.amazon.com/ses/latest/APIReference/API_ListIdentities.html
- aws_ses_email_identity table with pre applied filter: https://github.com/turbot/steampipe-plugin-aws/blob/main/aws/table_aws_ses_email_identity.go#L90

janritter:
# Integration test logs
N/A

# Example query results
<details>
  <summary>Results</summary>
  
```
select * from aws.aws_ses_domain_identity limit 1;

 +-------------------------------+------------------------------------------------------------------------------+---------------------+----------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-------------------------------+----------------------------------------------------------------------------------+-----------+--------------+--------------+---------------------------+
| identity                      | arn                                                                          | verification_status | verification_token                           | notification_attributes                                                                                                                                                                                                           | title                         | akas                                                                             | partition | region       | account_id   | _ctx                      |
+-------------------------------+------------------------------------------------------------------------------+---------------------+----------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-------------------------------+----------------------------------------------------------------------------------+-----------+--------------+--------------+---------------------------+
| redacted.example.com          | arn:aws:ses:eu-central-1:1234567890:identity/redacted.example.com            | Success             | redacted                                     | {"BounceTopic":null,"ComplaintTopic":null,"DeliveryTopic":null,"ForwardingEnabled":true,"HeadersInBounceNotificationsEnabled":false,"HeadersInComplaintNotificationsEnabled":false,"HeadersInDeliveryNotificationsEnabled":false} | redacted.example.com          | ["arn:aws:ses:eu-central-1:1234567890:identity/redacted.example.com"]            | aws       | eu-central-1 | 1234567890   | {"connection_name":"aws"} |
+-------------------------------+------------------------------------------------------------------------------+---------------------+----------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-------------------------------+----------------------------------------------------------------------------------+-----------+--------------+--------------+---------------------------+
```
</details>

### DIFF
--- a/aws/plugin.go
+++ b/aws/plugin.go
@@ -315,6 +315,7 @@ func Plugin(ctx context.Context) *plugin.Plugin {
 			"aws_servicequotas_service_quota":                              tableAwsServiceQuotasServiceQuota(ctx),
 			"aws_servicequotas_service_quota_change_request":               tableAwsServiceQuotasServiceQuotaChangeRequest(ctx),
 			"aws_ses_email_identity":                                       tableAwsSESEmailIdentity(ctx),
+			"aws_ses_domain_identity":                                      tableAwsSESDomainIdentity(ctx),
 			"aws_sfn_state_machine":                                        tableAwsStepFunctionsStateMachine(ctx),
 			"aws_sfn_state_machine_execution":                              tableAwsStepFunctionsStateMachineExecution(ctx),
 			"aws_sfn_state_machine_execution_history":                      tableAwsStepFunctionsStateMachineExecutionHistory(ctx),
--- a/aws/table_aws_ses_domain_identity.go
+++ b/aws/table_aws_ses_domain_identity.go
@@ -1,0 +1,186 @@
+package aws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ses"
+	"github.com/turbot/go-kit/types"
+	"github.com/turbot/steampipe-plugin-sdk/v3/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v3/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v3/plugin/transform"
+)
+
+func tableAwsSESDomainIdentity(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_ses_domain_identity",
+		Description: "AWS SES Domain Identity",
+		List: &plugin.ListConfig{
+			Hydrate: listSESDomainIdentities,
+		},
+		GetMatrixItem: BuildRegionList,
+		Columns: awsRegionalColumns([]*plugin.Column{
+			{
+				Name:        "identity",
+				Description: "The domain identity.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromValue(),
+			},
+			{
+				Name:        "arn",
+				Description: "The ARN of the AWS SES identity.",
+				Type:        proto.ColumnType_STRING,
+				Hydrate:     getDomainIdentityARN,
+				Transform:   transform.FromValue(),
+			},
+			{
+				Name:        "verification_status",
+				Description: "The verification status of the identity.",
+				Type:        proto.ColumnType_STRING,
+				Hydrate:     getDomainIdentityVerificationAttributes,
+				Transform:   transform.FromField("VerificationStatus"),
+			},
+			{
+				Name:        "verification_token",
+				Description: "The verification token for a domain identity.",
+				Type:        proto.ColumnType_STRING,
+				Hydrate:     getDomainIdentityVerificationAttributes,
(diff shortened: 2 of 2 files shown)


aws/plugin.go
```diff
@@ -315,6 +315,7 @@ func Plugin(ctx context.Context) *plugin.Plugin {
 			"aws_servicequotas_service_quota":                              tableAwsServiceQuotasServiceQuota(ctx),
 			"aws_servicequotas_service_quota_change_request":               tableAwsServiceQuotasServiceQuotaChangeRequest(ctx),
 			"aws_ses_email_identity":                                       tableAwsSESEmailIdentity(ctx),
+			"aws_ses_domain_identity":                                      tableAwsSESDomainIdentity(ctx),
```
cbruno10, line 318:
```suggestion
			"aws_ses_domain_identity":                                      tableAwsSESDomainIdentity(ctx),
			"aws_ses_email_identity":                                       tableAwsSESEmailIdentity(ctx),
```

Can you please sort in alphabetical order?

aws/table_aws_ses_domain_identity.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// Create Session
+	svc, err := SESService(ctx, d, region)
+	if err != nil {
+		return nil, err
+	}
+
+	input := &ses.ListIdentitiesInput{
+		MaxItems:     aws.Int64(1000),
+		IdentityType: aws.String(ses.IdentityTypeDomain),
```
cbruno10, line 91: I didn't know they had these constants, but thanks for using it/letting us know, we'll update the other table to use the email equivalent as well.

aws/table_aws_ses_domain_identity.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+			return !lastPage
+		},
+	)
+	return nil, err
+}
+
+//// HYDRATE FUNCTIONS
+
+func getDomainIdentityVerificationAttributes(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
```
cbruno10, line 126: For this hydrate function and others, the code looks similar/identical to those defined for the `aws_ses_email_identity` table.

I think in this case, we can re-use the hydrate functions from `table_aws_ses_email_identity.go`. and can also rename them as well to be more generic, e.g., `getEmailIdentityVerificationAttributes` -> `getSESIdentityVerificationAttributes`, `getEmailIdentityNotificationAttributes` -> `getSESIdentityNotificationAttributes`.

In this table file, then the only change you'd need to make is updating references to those files, e.g.,
```
{
	Name:        "verification_token",
	Description: "The verification token for a domain identity.",
	Hydrate:     getSESIdentityVerificationAttributes,
	Transform:   transform.FromField("VerificationToken"),
},
```

cbruno10 (changes requested):
@janritter Thanks for submitting this PR! The table code looks great, I've added a few minor suggestions.

Can you please also add a table doc for this new table with query examples? Please see `docs/tables/aws_ses_email_identity.md` for an example.

aws/table_aws_ses_domain_identity.go
janritter: I also updated the ses_email_identity table to use the constant

aws/table_aws_ses_domain_identity.go
janritter: Good point, I was actually thinking about this when copying the code from email_identity, but not sure if tables should "reuse" hydrate functions between them. Renamed the hydrate functions and updated both tables.

janritter: Good point, I was actually thinking about this when copying the code from email_identity, but not sure if tables should "reuse" hydrate functions between them. Renamed the hydrate functions and updated both tables.

cbruno10 (approved):

cbruno10:
@janritter Thanks for the PR and the quick updates after the review! We'll release this table later today.

PR #1285, opened 2022-09-02
Issue 1262

niharika2k00:
Fix column `url_config` in table aws_lambda_function to return nil incase of `aws gov cloud.`


aws/table_aws_lambda_function.go
```diff
@@ -362,6 +362,17 @@ func getLambdaFunctionUrlConfig(ctx context.Context, d *plugin.QueryData, h *plu
 
 	functionName := functionName(h.Item)
 
+	commonColumnData, err := getCommonColumns(ctx, d, h)
+	if err != nil {
+		return nil, err
+	}
+
+	awsCommonData := commonColumnData.(*awsCommonColumnData)
+	if awsCommonData.Partition == "aws-us-gov"{
```
cbruno10, line 371:
```suggestion
	if awsCommonData.Partition == "aws-us-gov" {
```

@niharika2k00 can you please either run `gofmt` on your code to format it, or ensure that it's setup in your VS Code so it automatically runs?

cbruno10, line 371: Can you please add a comment above the `if` statement stating why we need this special handling? The comment should link to the AWS article mentioning limitations in GovCloud and the error we get (please see the original issue for the article link)

aws/table_aws_lambda_function.go
```diff
@@ -362,6 +362,17 @@ func getLambdaFunctionUrlConfig(ctx context.Context, d *plugin.QueryData, h *plu
...
+		plugin.Logger(ctx).Debug("getLambdaFunctionUrlConfig", "aws-gov-cloud encountered", functionName)
```
cbruno10, line 372:
```suggestion
```

I think it's ok if we don't have this logging statement, since the return is straightforward

cbruno10 (changes requested):
@niharika2k00 Please see review comments, thanks!

cbruno10 (approved):

PR #1315, opened 2022-09-26
feat: add ecr image finding table

aminvielledebatAtBedrock:
This table is used to get findings on a specific tag, in a specific repository allowing joins with others data sources

# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example ecr image tag findings

``` sql
select 
    repository_name
    ,image_tag
    ,image_scan_completed_at
    ,vulnerability_source_updated_at
    ,finding
    ,partition
    ,repository_name
    ,account_id
from services.aws_ecr_image_findings where repository_name = 'my-ecr-repository-name' and image_tag = 'v1.0.0'
```

# Get findings on images tags pushed in the last 24 hours

```sql
select
    finding.*
from (
    select
        repository_name
        ,jsonb_array_elements_text(image_tags) as image_tag
    from services.aws_ecr_image i
    where
        i.image_pushed_at > now() - interval '24' hour
) images
left outer join services.aws_ecr_image_findings finding
on images.repository_name = finding.repository_name and images.image_tag = finding.image_tag
```

### DIFF
--- a/aws/plugin.go
+++ b/aws/plugin.go
@@ -164,6 +164,7 @@ func Plugin(ctx context.Context) *plugin.Plugin {
 			"aws_ec2_transit_gateway_route_table":                          tableAwsEc2TransitGatewayRouteTable(ctx),
 			"aws_ec2_transit_gateway_vpc_attachment":                       tableAwsEc2TransitGatewayVpcAttachment(ctx),
 			"aws_ecr_image":                                                tableAwsEcrImage(ctx),
+			"aws_ecr_image_scan_finding":                                   tableAwsEcrImageScanFinding(ctx),
 			"aws_ecr_repository":                                           tableAwsEcrRepository(ctx),
 			"aws_ecrpublic_repository":                                     tableAwsEcrpublicRepository(ctx),
 			"aws_ecs_cluster":                                              tableAwsEcsCluster(ctx),
--- a/aws/table_aws_ecr_image_scan_finding.go
+++ b/aws/table_aws_ecr_image_scan_finding.go
@@ -1,0 +1,168 @@
+package aws
+
+import (
+	"context"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/turbot/steampipe-plugin-sdk/v4/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v4/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v4/plugin/transform"
+	"time"
+)
+
+//// TABLE DEFINITION
+
+func tableAwsEcrImageScanFinding(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_ecr_image_scan_finding",
+		Description: "AWS ECR Image Scan findings",
+		List: &plugin.ListConfig{
+			Hydrate: getAwsEcrImageScanFindings,
+			IgnoreConfig: &plugin.IgnoreConfig{
+				ShouldIgnoreErrorFunc: isNotFoundError([]string{"RepositoryNotFoundException", "ImageNotFoundException", "ScanNotFoundException"}),
+			},
+			KeyColumns: []*plugin.KeyColumn{
+				{Name: "repository_name", Require: plugin.Required},
+				{Name: "image_tag", Require: plugin.Required},
+			},
+		},
+		GetMatrixItemFunc: BuildRegionList,
+		Columns: awsRegionalColumns([]*plugin.Column{
+			{
+				Name:        "repository_name",
+				Description: "The name of the repository.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "image_tag",
+				Description: "The image tag",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "image_digest",
+				Description: "The image digest",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "image_scan_status",
+				Description: "The current state of the scan",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "image_scan_completed_at",
+				Description: "The date and time, in JavaScript date format, when the repository was created.",
+				Type:        proto.ColumnType_TIMESTAMP,
+			},
+			{
+				Name:        "vulnerability_source_updated_at",
+				Description: "The date and time, in JavaScript date format, when the repository was created.",
+				Type:        proto.ColumnType_TIMESTAMP,
+			},
+			{
+				Name:        "name",
+				Description: "The name associated with the finding, usually a CVE number.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("ImageScanFinding.Name"),
+			},
+			{
+				Name:        "severity",
+				Description: "The finding severity.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("ImageScanFinding.Severity"),
+			},
+			{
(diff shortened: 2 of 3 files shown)


cbruno10:
Hey @aminvielledebatAtBedrock , thanks for working on and opening this PR!

Are you ready for us to review the PR yet, or are you still planning on pushing in some changes (since the PR is in draft mode)?

aminvielledebatAtBedrock:
Nope @cbruno10. It needs unit tests and documentation.
I opened it in draft to have your point of view about this new table.

It will return N rows for N findings for a specific image tag, and allow us to join rows with images from our kubernetes clusters without aws rate limiting (whereas `table_aws_ecr_repository` returns all findings for all images tags).

If you're interested about it, i can add documentation and tests quickly.

cbruno10:
@aminvielledebatAtBedrock I think this new table is a great idea, I actually didn't even realize we return all image scan findings in the `image_scanning_findings` column in the `aws_ecr_repository` table. We actually made a similar change recently in that same table, where we had a column `image_details`, but we decided to create the `aws_ecr_image` table and plan on deprecating the `image_details` column actually.

One design aspect we're considering for this table is should it require a user to specify the image ID/tag and the repository (like they are in this PR), or should we leverage the `ParentHydrate` function, which allows us to re-use functions to list parent resources, eliminating the need for required key columns.

For instance, in the `aws_ecr_image` table, we use a [ParentHydrate](https://github.com/turbot/steampipe-plugin-aws/blob/a8da808552dd4d31c1326285499f3396d7866b3a/aws/table_aws_ecr_image.go#L19) to list all repositories first, and then list images in each repository. If a user passes in a specific repository name though, it'll only get images in that repository.

In this case though, it'd potentially be 3 levels of nesting though, e.g., Repository -> Image -> Scan Findings, which is a lot, and like you mentioned, may lead to throttling.

If you're able to share, approximately how many repositories, images, and findings do you have in your environments where you'd be using this table? And do most of your accounts/environments have a similar number, or are there varied amounts?

aminvielledebatAtBedrock:
No problem !

We have :  
~250 images deployed in production (even more in other environments / tenants...). These production images are stored in only one service account.
~200 repositories for production in this account (~700 with non production)
~ an average of about 80 findings pointed by ecr scan (when you take account low and informational findings)
~ an ecr lifecycle policy wich delete oldest images when you have more than 300 image tags in your repository 

So we have 200 x 300 x 80 = 480 0000  findings record only for production :disappointed: 

`ParentHydrate` doesn't work in that case :'(

aws/table_aws_ecr_image_findings.go
```diff
@@ -0,0 +1,134 @@
...
+				Description: "Finding",
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("Finding"),
+			},
+		}),
+	}
+}
+
+// // LIST FUNCTION
+func getAwsEcrImageScanningFindings(ctx context.Context, d *plugin.QueryData, _ *plugin.HydrateData) (interface{}, error) {
```
villers, line 63:
```suggestion
func getAwsEcrImageScanningFindings(ctx context.Context, d *plugin.QueryData, _ *plugin.HydrateData) (any, error) {
```

villers: im note sure with this :p

aws/table_aws_ecr_image_findings.go
```diff
@@ -0,0 +1,134 @@
...
+	input := &ecr.DescribeImageScanFindingsInput{
+		MaxResults:     aws.Int64(1000),
+		RepositoryName: aws.String(repositoryName.GetStringValue()),
+		ImageId: &ecr.ImageIdentifier{
+			ImageTag: aws.String(imageTag.GetStringValue()),
+		},
+	}
+
+	limit := d.QueryContext.Limit
+	if d.QueryContext.Limit != nil {
```
villers, line 88:
```suggestion
	limit := d.QueryContext.Limit
	if limit != nil {
```

aws/table_aws_ecr_image_findings.go
aminvielledebatAtBedrock: It's the same in the codebase. So.... don't fix if it works :)
Thanks @villers

villers (approved):

cbruno10:
Thanks @aminvielledebatAtBedrock for the info!

Agree that trying to list that many image findings isn't really scalable, and even if we did get through all of the throttling exceptions, it'd take quite a bit of time to get all of the results.

With that in mind, I think requiring the repository and image digest or tag makes sense. I have a few other suggestions/questions for the table as well:
- Can you please update the table name and filename to be `aws_ecr_image_scan_finding`? Usually our table names are singular, and I think adding the word `scan` makes sense, as the API + types include that word as well.
- Since the API also allows a user to pass in the image digest, can you please add a new column and optional qual `image_digest`? In the list call, if neither an image digest or tag was passed, the function should return an error.
- Looking at the [response syntax](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_DescribeImageScanFindings.html#API_DescribeImageScanFindings_ResponseSyntax) for the API, there's the `imageScanStatus` field, was there a reason this was excluded as a column (`image_scan_status`)?
- Instead of having a single column `finding`, I think it'd be better if each field from the [ImageScanFinding type](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_ImageScanFinding.html) was its own column, e.g., `attributes`, `description`, `name`, etc.
- The other columns, like `image_scan_completed_at` and `vulnerability_source_updated_at` would still be columns as well.
- Also from the response syntax, I can see there are two sets of findings, `enhanceFindings` and `findings`. Both type of finding seems like there's a sufficient number of columns to be their own tables, e.g., `aws_ecr_image_scan_finding`, `aws_ecr_image_enhanced_scan_finding`. Does this also seem reasonable to you, based on your knowledge of enhanced findings vs. findings?

aminvielledebatAtBedrock:
Hi @cbruno10 !

I did the changes on the `aws_ecr_image_scan_finding` table , excepting one : allowing usage of `image_digest` as a new optional qual.
I'm not able to use it when i join results from another table... I spend a lot of time without finding any answer to my issue.
Maybe you've got a good idea...

The `plugin.HydrateData` parameter is always empty (as expected because we don't have a `parentHydrate`), `plugin.QueryData.KeyColumnQuals` is always empty too on Optional fields `image_digest` and `image_tag` with the following queries : 

``` sql
select
  images.repository_name,
  images.image_tag,
  findings.* 
from
  (
    select
      repository_name,
      jsonb_array_elements_text(image_tags) as image_tag 
    from
      services.aws_ecr_image i 
    where
      i.image_pushed_at > now() - interval '24' hour 
  )
  images 
  inner join
    services.aws_ecr_image_scan_finding findings 
    on images.repository_name = findings.repository_name 
    and images.image_tag = findings.image_tag
```
Returns `Error: Ecr findings called without repositoryName or imageTag or imageDigest (SQLSTATE HV000)`

``` sql
select
  * 
from
  services.aws_ecr_image_scan_finding f 
where
  (
    f.repository_name,
    f.image_tag
  )
  in 
  (
    select
      repository_name,
      image_tag 
    from
      services.aws_ecr_image i 
    where
      i.image_pushed_at > now() - interval '24' hour 
    group by
      1,
      2 
  )
```
returns `Error: rpc error: code = Internal desc = 'List' call for table 'aws_ecr_image_scan_finding' is missing 1 required qual: column:'repository_name' operator: =
 (SQLSTATE HV000)` 

If you have any suggestions please advise me :pray:

cbruno10 (approved):

cbruno10:
Thanks @aminvielledebatAtBedrock for making the changes so quickly and adding this table!

PR #1324, opened 2022-10-07
Update plugin to use github.com/aws/aws-sdk-go-v2 - Week 5

LalitTurbot:


aws/utils.go
```diff
@@ -225,6 +226,11 @@ func handleEmptySliceAndMap(ctx context.Context, d *transform.TransformData) (an
 		if reflectVal == reflect.Zero(reflectVal.Type()) {
 			return nil, nil
 		}
+	case reflect.String:
+		// To handle empty string to null change in aws sdk V1 to V2 migration
```
cbruno10, line 230: @LalitTurbot Do we still need this since we're not doing v1 backward compatible changes now?

aws/service_v2.go
```diff
@@ -204,17 +295,38 @@ func CostExplorerClient(ctx context.Context, d *plugin.QueryData) (*costexplorer
...
+	cfg, err := getClientForQueryRegion(ctx, d)
 	if err != nil {
 		return nil, err
 	}
-	if cfg == nil {
-		return nil, nil
+	return databasemigrationservice.NewFromConfig(*cfg), nil
+}
+
+func DaxClient(ctx context.Context, d *plugin.QueryData) (*dax.Client, error) {
```
cbruno10, line 306:
```suggestion
func DAXClient(ctx context.Context, d *plugin.QueryData) (*dax.Client, error) {
```

aws/service_v2.go
```diff
@@ -267,6 +423,39 @@ func ELBV2Client(ctx context.Context, d *plugin.QueryData) (*elasticloadbalancin
...
+	if err != nil {
+		return nil, err
+	}
+	if cfg == nil {
+		return nil, nil
+	}
+	return efs.NewFromConfig(*cfg), nil
+}
+
+func FsxClient(ctx context.Context, d *plugin.QueryData) (*fsx.Client, error) {
```
cbruno10, line 437:
```suggestion
func FSXClient(ctx context.Context, d *plugin.QueryData) (*fsx.Client, error) {
```

cbruno10 (changes requested):
@LalitTurbot Please see comments, thanks!

aws/utils.go
LalitTurbot: No, we don't need it now. But I have used it in some tables from week2. So did not remove it yet.
Also added handling for the string in it. Will remove it uses further if any in week 5 PR

aws/service_v2.go
LalitTurbot: @cbruno10, should it be `FSx` or `FSX`?
In documents, it is `FSx`.
https://docs.aws.amazon.com/fsx/latest/WindowsGuide/what-is.html

cbruno10: `FSx` is correct, thanks for the correction!

cbruno10 (approved):

PR #1369, opened 2022-10-28
feat: add aws_pricing_product table

aminvielledebatAtBedrock:
# Context

Allowing usage of aws pricing API

</details>

# Example query results
<details>
  <summary>Results</summary>

```
select
    service_code,
    instance_type,
    cache_engine,
    term,
    purchase_option,
    lease_contract_length,
    description,
    price_per_unit,
    currency
from
    aws_pricing_product
where
    service_code = 'AmazonElastiCache'
  and region_code = 'eu-west-3'
  and location_type = 'AWS Region'
  and instance_type = 'cache.m5.xlarge'
  and cache_engine = 'Redis';
```

| service\_code     | instance\_type  | cache\_engine | term     | purchase\_option | lease\_contract\_length | description                                             | price\_per\_unit | currency |
| ----------------- | --------------- | ------------- | -------- | ---------------- | ----------------------- | ------------------------------------------------------- | ---------------- | -------- |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | No Upfront       | 3yr                     | Redis, cache.m5.xlarge reserved instance applied        | 0.1870000000     | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | OnDemand |                  |                         | $0.36 per Gen Purpose M5.xlarge node hour running Redis | 0.3600000000     | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | Partial Upfront  | 1yr                     | Redis, cache.m5.xlarge reserved instance applied        | 0.1170000000     | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | All Upfront      | 1yr                     | Upfront Fee                                             | 2009             | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | All Upfront      | 3yr                     | Redis, cache.m5.xlarge reserved instance applied        | 0.0000000000     | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | No Upfront       | 1yr                     | Redis, cache.m5.xlarge reserved instance applied        | 0.2460000000     | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | All Upfront      | 1yr                     | Redis, cache.m5.xlarge reserved instance applied        | 0.0000000000     | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | Partial Upfront  | 3yr                     | Redis, cache.m5.xlarge reserved instance applied        | 0.0860000000     | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | All Upfront      | 3yr                     | Upfront Fee                                             | 4269             | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | Partial Upfront  | 3yr                     | Upfront Fee                                             | 2271             | USD      |
| AmazonElastiCache | cache.m5.xlarge | Redis         | Reserved | Partial Upfront  | 1yr                     | Upfront Fee                                             | 1025             | USD      |

</details>

### DIFF
--- a/aws/table_aws_pricing_product.go
+++ b/aws/table_aws_pricing_product.go
@@ -216,8 +216,11 @@ func extractCurrency(_ context.Context, d *transform.TransformData) (interface{}
 
 // build pricing list call input filter
 func buildPricingFilter(qual *plugin.KeyColumnQuals) ([]types.Filter, error) {
-	filters := make([]types.Filter, 0)
+	if qual == nil {
+		return nil, nil
+	}
 
+	filters := make([]types.Filter, 0)
 	for _, qual := range qual.Quals {
 		qualFilter := make(map[string]*string)
 		err := json.Unmarshal([]byte(qual.Value.GetJsonbValue()), &qualFilter)
--- a/aws/table_aws_pricing_product.go
+++ b/aws/table_aws_pricing_product.go
@@ -1,0 +1,238 @@
+package aws
+
+import (
+	"context"
+	"encoding/json"
+	"github.com/aws/aws-sdk-go-v2/service/pricing"
+	"github.com/aws/aws-sdk-go-v2/service/pricing/types"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/turbot/steampipe-plugin-sdk/v4/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v4/plugin/transform"
+	"time"
+
+	"github.com/turbot/steampipe-plugin-sdk/v4/plugin"
+)
+
+//// TABLE DEFINITION
+
+func tableAwsPricingProduct(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_pricing_product",
+		Description: "AWS Pricing Product",
+		List: &plugin.ListConfig{
+			Hydrate: listPricingProduct,
+			KeyColumns: []*plugin.KeyColumn{
+				{Name: "service_code", Require: plugin.Required},
+				{Name: "filters", Require: plugin.Optional, CacheMatch: "exact"},
+			},
+		},
+		Columns: awsDefaultColumns([]*plugin.Column{
+			{Name: "rate_code", Description: "A unique code for a product/ offer/ pricing-tier combination.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.RateCode")},
+			{Name: "service_code", Description: "This identifies the specific AWS service to the customer as a unique short abbreviation.", Type: proto.ColumnType_STRING, Transform: transform.FromField("ServiceCode")},
+			{Name: "term", Description: "Whether your AWS usage is Reserved or On-Demand.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.Term")},
+			{Name: "purchase_option", Description: "How you chose to pay for this line item (All Upfront, Partial Upfront, No Upfront).", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.TermAttributes.PurchaseOption")},
+			{Name: "lease_contract_length", Description: "The length of time that your RI is reserved for.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.TermAttributes.LeaseContractLength")},
+			{Name: "description", Description: "Description for a product / offer / pricing-tier combination.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.Description")},
+			{Name: "begin_range", Description: "Start of billing range, by unit", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.BeginRange")},
+			{Name: "end_range", Description: "Enf of billing range, by unit", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.EndRange")},
+			{Name: "unit", Description: "The pricing unit that AWS used for calculating your usage cost (ex: hours)", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.Unit")},
+			{Name: "price_per_unit", Description: "Price by unit", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.PricePerUnit").Transform(extractPricePerUnit)},
+			{Name: "currency", Description: "Currency used for the price", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.PricePerUnit").Transform(extractCurrency)},
+			{Name: "publication_date", Description: "The publication date of the offer.", Type: proto.ColumnType_TIMESTAMP, Transform: transform.FromField("PublicationDate")},
+			{Name: "effective_date", Description: "The effective date of the pricing details.", Type: proto.ColumnType_TIMESTAMP, Transform: transform.FromField("Offer.EffectiveDate")},
+			{Name: "version", Description: "The publication version of the offer.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Version")},
+			// product attributes
+			{Name: "attributes", Description: "Product attributes.", Type: proto.ColumnType_JSON, Transform: transform.FromField("Product.Attributes")},
+			// product attributes filters
+			{Name: "filters", Description: "Product filtering by attribute.", Type: proto.ColumnType_JSON, Transform: transform.FromQual("filters")},
+		}),
+	}
+}
+
+//// LIST FUNCTION
+
+type Product struct {
+	_             struct{} `type:"structure"`
+	ProductFamily *string
+	Attributes    map[string]*string
+}
+
+type Term map[string]*Offer
+
+type PriceDimension struct {
+	Unit         *string
+	BeginRange   *string
+	EndRange     *string
+	Description  *string
+	RateCode     *string
+	PricePerUnit map[string]*string
+}
+
+type TermAttributes struct {
+	_                   struct{} `type:"structure"`
+	PurchaseOption      *string
+	LeaseContractLength *string
+}
+
+type Offer struct {
+	_               struct{} `type:"structure"`
+	PriceDimensions map[string]*PriceDimension
+	EffectiveDate   *time.Time
+	OfferTermCode   *string
+	TermAttributes  *TermAttributes
+}
+
+type PriceList struct {
+	_               struct{} `type:"structure"`
+	Product         *Product
+	ServiceCode     *string
+	Terms           map[string]*Term
+	Version         *string
+	PublicationDate *time.Time
+}
+
+func listPricingProduct(ctx context.Context, d *plugin.QueryData, _ *plugin.HydrateData) (interface{}, error) {
+	// Create Session
+	svc, err := PricingServiceClient(ctx, d)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_pricing_product.listPricingProduct", "connection_error", err)
+		return nil, err
+	}
+	if svc == nil {
+		// Unsupported region, return no data
+		return nil, nil
+	}
+
+	// Limiting the results
+	maxItems := int32(100)
+	if d.QueryContext.Limit != nil {
+		limit := int32(*d.QueryContext.Limit)
+		if limit < maxItems {
+			if limit < 20 {
+				maxItems = 20
+			} else {
+				maxItems = limit
+			}
+		}
+	}
+
+	equalQual := d.KeyColumnQuals
+	input := pricing.GetProductsInput{
+		ServiceCode:   aws.String(equalQual["service_code"].GetStringValue()),
+		FormatVersion: aws.String("aws_v1"),
+		MaxResults:    maxItems,
+	}
+
+	type OfferOutput struct {
+		_              struct{} `type:"structure"`
+		PriceDimension *PriceDimension
+		Term           string
+		EffectiveDate  *time.Time
+		OfferTermCode  *string
+		TermAttributes *TermAttributes
+	}
+
+	type PriceOutput struct {
+		_               struct{} `type:"structure"`
+		Product         *Product
+		ServiceCode     *string
+		Offer           *OfferOutput
+		Version         *string
+		PublicationDate *time.Time
+	}
+
+	filters, err := buildPricingFilter(d.Quals["filters"])
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_pricing_product.listPricingProduct", "filters_building_error", err)
+		return nil, err
+	}
+
+	if len(filters) > 0 {
+		input.Filters = filters
+	}
+
+	paginator := pricing.NewGetProductsPaginator(svc, &input, func(o *pricing.GetProductsPaginatorOptions) {
+		o.Limit = maxItems
+		o.StopOnDuplicateToken = true
+	})
+
+	for paginator.HasMorePages() {
+		output, err := paginator.NextPage(ctx)
+		if err != nil {
+			plugin.Logger(ctx).Error("aws_pricing_product.listPricingProduct", "api_error", err)
+			return nil, err
+		}
+
+		for _, priceList := range output.PriceList {
+			var priceListObject PriceList
+
+			err = json.Unmarshal([]byte(priceList), &priceListObject)
+			if err != nil {
+				plugin.Logger(ctx).Error("aws_pricing_product.listPricingProduct", "unmarshal_err", err)
+				return nil, err
+			}
+
+			for term, termDetail := range priceListObject.Terms {
+				for _, offer := range *termDetail {
+					for _, priceDimension := range offer.PriceDimensions {
+						d.StreamListItem(ctx, PriceOutput{
+							Product:         priceListObject.Product,
+							ServiceCode:     priceListObject.ServiceCode,
+							Version:         priceListObject.Version,
+							PublicationDate: priceListObject.PublicationDate,
+							Offer:           &OfferOutput{PriceDimension: priceDimension, Term: term, EffectiveDate: offer.EffectiveDate, OfferTermCode: offer.OfferTermCode, TermAttributes: offer.TermAttributes},
+						})
+					}
+				}
+			}
+
+			// Context may get cancelled due to manual cancellation or if the limit has been reached
+			if d.QueryStatus.RowsRemaining(ctx) == 0 {
+				return nil, nil
+			}
+		}
+	}
+
+	return nil, nil
+}
+
+func extractPricePerUnit(_ context.Context, d *transform.TransformData) (interface{}, error) {
+	priceMap := d.Value.(map[string]*string)
+
+	for _, units := range priceMap {
+		return units, nil
+	}
+	return nil, nil
+}
+
+func extractCurrency(_ context.Context, d *transform.TransformData) (interface{}, error) {
+	priceMap := d.Value.(map[string]*string)
+
+	for currency := range priceMap {
+		return currency, nil
+	}
+	return nil, nil
+}
+
+// build pricing list call input filter
+func buildPricingFilter(qual *plugin.KeyColumnQuals) ([]types.Filter, error) {
+	filters := make([]types.Filter, 0)
+
+	for _, qual := range qual.Quals {
+		qualFilter := make(map[string]*string)
+		err := json.Unmarshal([]byte(qual.Value.GetJsonbValue()), &qualFilter)
+		if err != nil {
+			return nil, err
+		}
+
+		for attributeName, attributeValue := range qualFilter {
+			filter := types.Filter{
+				Field: aws.String(attributeName),
+				Type:  types.FilterTypeTermMatch,
+			}
+			filter.Value = attributeValue
+			filters = append(filters, filter)
+		}
+	}
+	return filters, nil
+}
--- a/docs/tables/aws_pricing_product.md
+++ b/docs/tables/aws_pricing_product.md
@@ -1,0 +1,91 @@
+# Table: aws_pricing_product
+
+Returns costs, and offer definition, for the corresponding service.
+
+## Examples
+
+### List pricing offers for on-demand shared EC2 c5.2xlarge without pre-installed software, with Linux OS
+
+```sql
+select
+  term,
+  purchase_option,
+  lease_contract_length,
+  unit,
+  price_per_unit::numeric::money,
+  currency,
+  begin_range,
+  end_range,
+  effective_date,
+  description,
+  attributes ->> 'instanceType',
+  attributes ->> 'vcpu',
+  attributes ->> 'memory',
+  attributes ->> 'operatingSystem',
+  attributes ->> 'preInstalledSw'
+from
+  aws_pricing_product 
+where
+  service_code = 'AmazonEC2' 
+  and filters = '{
+  "regionCode": "eu-west-3",
+  "locationType": "AWS Region",
+  "instanceType": "c5.2xlarge",
+  "operatingSystem": "Linux",
+  "tenancy": "Shared",
+  "preInstalledSw": "NA",
+  "capacityStatus": "Used" }'::jsonb;
+```
+
+
+### List pricing offers for Mysql RDS db.m5.xlarge instance in eu-west-3 in single-az deployment
+
+```sql
+select
+  term,
+  purchase_option,
+  lease_contract_length,
+  unit,
+  price_per_unit::numeric::money,
+  currency,
+  attributes ->> 'instanceType',
+  attributes ->> 'vcpu',
+  attributes ->> 'memory',
+  attributes ->> 'databaseEngine',
+  attributes ->> 'deploymentOption' 
+from
+  aws_pricing_product 
+where
+  service_code = 'AmazonRDS' 
+  and filters = '{
+  "regionCode": "eu-west-3",
+  "locationType": "AWS Region",
+  "instanceType": "db.m5.xlarge",
+  "databaseEngine": "MySQL",
+  "deploymentOption": "Single-AZ" }'::jsonb;
+```
+
+### List pricing offers for Redis ElasticCache cache.m5.xlarge instances in eu-west-3
+
+```sql
+select
+  term,
+  purchase_option,
+  lease_contract_length,
+  unit,
+  price_per_unit::numeric::money,
+  currency,
+  attributes ->> 'instanceType',
+  attributes ->> 'vcpu',
+  attributes ->> 'memory',
+  attributes ->> 'cacheEngine' 
+from
+  aws_pricing_product 
+where
+  service_code = 'AmazonElastiCache' 
+  and filters = '{
+  "regionCode": "eu-west-3",
+  "locationType": "AWS Region",
+  "instanceType": "cache.m5.xlarge",
+  "cacheEngine": "Redis" }'::jsonb;
+```
--- a/aws/plugin.go
+++ b/aws/plugin.go
@@ -274,6 +274,7 @@ func Plugin(ctx context.Context) *plugin.Plugin {
 			"aws_opensearch_domain":                                        tableAwsOpenSearchDomain(ctx),
 			"aws_organizations_account":                                    tableAwsOrganizationsAccount(ctx),
 			"aws_pinpoint_app":                                             tableAwsPinpointApp(ctx),
+			"aws_pricing_product":                                          tableAwsPricingProduct(ctx),
 			"aws_pricing_service_attribute":                                tableAwsPricingServiceAttribute(ctx),
 			"aws_ram_principal_association":                                tableAwsRAMPrincipalAssociation(ctx),
 			"aws_ram_resource_association":                                 tableAwsRAMResourceAssociation(ctx),
--- a/aws/service.go
+++ b/aws/service.go
@@ -852,7 +852,10 @@ func PinpointClient(ctx context.Context, d *plugin.QueryData) (*pinpoint.Client,
 }
 
 func PricingServiceClient(ctx context.Context, d *plugin.QueryData) (*pricing.Client, error) {
-	cfg, err := getClient(ctx, d, getDefaultAwsRegion(d))
+	// Pricing API is a global API that supports only us-east-1 and ap-south-1 regions
+	// getDefaultAwsRegion doesn't return the good region at the moment (it should use specified API endpoints but it doesn't).
+	// Set us-east-1 for now
+	cfg, err := getClient(ctx, d, "us-east-1")
 	if err != nil {
 		return nil, err
 	}


aminvielledebatAtBedrock:
cc @cbruno10 with another one . It's ready for me :-)

docs/tables/aws_pricing_product.md
```diff
@@ -0,0 +1,55 @@
...
+  aws_pricing_product 
+where
+  service_code = 'AmazonEC2' 
+  and region_code = 'eu-west-3' 
+  and location_type = 'AWS Region' 
+  and instance_type = 'c5.2xlarge' 
+  and operating_system = 'Linux' 
+  and tenancy = 'Shared' 
+  and pre_installed_software = 'NA' 
+  and capacity_status = 'Used'
```
misraved, line 22:
```suggestion
  and capacity_status = 'Used';
```

docs/tables/aws_pricing_product.md
```diff
@@ -0,0 +1,55 @@
...
+  * 
+from
+  aws_pricing_product 
+where
+  service_code = 'AmazonRDS' 
+  and region_code = 'eu-west-3' 
+  and location_type = 'AWS Region' 
+  and instance_type = 'db.m5.xlarge' 
+  and database_engine = 'MySQL' 
+  and deployment_option = 'Single-AZ'
```
misraved, line 39:
```suggestion
  and deployment_option = 'Single-AZ';
```

docs/tables/aws_pricing_product.md
```diff
@@ -0,0 +1,55 @@
...
+select
+  * 
+from
+  aws_pricing_product 
+where
+  service_code = 'AmazonElastiCache' 
+  and region_code = 'eu-west-3' 
+  and location_type = 'AWS Region' 
+  and instance_type = 'cache.m5.xlarge'
+  and cache_engine = 'Redis'
```
misraved, line 54:
```suggestion
  and cache_engine = 'Redis';
```

aws/table_aws_pricing_product.go
```diff
@@ -0,0 +1,274 @@
...
+			{Name: "service_code", Description: "This identifies the specific AWS service to the customer as a unique short abbreviation.", Type: proto.ColumnType_STRING, Transform: transform.FromField("ServiceCode")},
+			{Name: "term", Description: "Whether your AWS usage is Reserved or On-Demand.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.Term")},
+			{Name: "purchase_option", Description: "How you chose to pay for this line item. Valid values are All Upfront, Partial Upfront, and No Upfront.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.TermAttributes.PurchaseOption")},
+			{Name: "lease_contract_length", Description: "The length of time that your RI is reserved for.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.TermAttributes.LeaseContractLength")},
+			{Name: "description", Description: "Description for a product / offer / pricing-tier combination", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.Description")},
```
misraved, line 41:
```suggestion
			{Name: "description", Description: "Description for a product / offer / pricing-tier combination.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.Description")},
```

aws/table_aws_pricing_product.go
```diff
@@ -0,0 +1,274 @@
...
+			{Name: "begin_range", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.BeginRange")},
+			{Name: "end_range", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.EndRange")},
```
misraved, line 43: Please add the column description for the `begin_range` and the `end_range` columns.

misraved:
Thanks @aminvielledebatAtBedrock for this table 👍 .

There are several columns that are missing descriptions. Could you also add the missing period `.` at the end of certain column descriptions? Could you also try to trim down a few descriptions since they seem quite lengthy? For instance, the description for `region_code` could be trimmed down to just `A Region is a physical location around the world where data centers are clustered.`. While we love having good descriptions for our columns, I think we could possibly trim down a few from a self-explanatory point of view.

aws/table_aws_pricing_product.go
```diff
@@ -0,0 +1,274 @@
...
+			{Name: "unit", Description: "The pricing unit that AWS used for calculating your usage cost. For example, the pricing unit for Amazon EC2 instance usage is in hours.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.Unit")},
+			{Name: "price", Description: "Price by unit", Type: proto.ColumnType_DOUBLE, Transform: transform.FromField("Offer.PriceDimension.PricePerUnit").Transform(extractPrice)},
```
cbruno10, line 45:
```suggestion
			{Name: "price", Description: "Price by unit", Type: proto.ColumnType_STRING, Transform: transform.FromField("Offer.PriceDimension.PricePerUnit").Transform(extractPrice)},
```

Can you please change this to a string? We prefer string over doubles as doubles are never exact, and AWS normally provides detailed price info, i.e., 3+ decimal places.

cbruno10: Then in the examples, we use `::numeric::money` Postgres conversions to show users how to handle conversions, e.g., https://hub.steampipe.io/plugins/turbot/aws/tables/aws_cost_by_account_daily#basic-info

aws/table_aws_pricing_product.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	return nil, nil
+}
+
+func extractPrice(_ context.Context, d *transform.TransformData) (interface{}, error) {
+	priceMap := d.Value.(map[string]*string)
+
+	for _, units := range priceMap {
+		price, err := strconv.ParseFloat(*units, 32)
```
cbruno10, line 230: What are some examples of prices we get back from the API? Based on the suggested change to the column type, I think we wouldn't need to cast anymore?

aws/table_aws_pricing_product.go
```diff
@@ -0,0 +1,274 @@
...
+			{Name: "current_generation", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.currentGeneration")},
+			{Name: "pre_installed_software", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.preInstalledSw")},
+			{Name: "processor_architecture", Description: "Describes your processor architecture.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.processorArchitecture")},
+			// multi az attributes
+			{Name: "deployment_option", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.deploymentOption")},
+			// RDS fields
+			{Name: "database_engine", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.databaseEngine")},
+			{Name: "min_volume_size", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.minVolumeSize")},
+			{Name: "maxVolumeSize", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.maxVolumeSize")},
```
cbruno10, line 75:
```suggestion
			{Name: "max_volume_size", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.maxVolumeSize")},
```

aws/table_aws_pricing_product.go
```diff
@@ -0,0 +1,274 @@
...
+		price, err := strconv.ParseFloat(*units, 32)
+		return price, err
+	}
+	return nil, nil
+}
+
+func extractCurrency(_ context.Context, d *transform.TransformData) (interface{}, error) {
+	priceMap := d.Value.(map[string]*string)
+
+	for currency := range priceMap {
```
cbruno10, line 239: Maybe I'm not understanding this code well, but why do we have a for loop and then return the first item we get?

docs/tables/aws_pricing_product.md
```diff
@@ -0,0 +1,55 @@
...
+
+Returns costs, and offer definition, for the corresponding service.
+
+## Examples
+
+### List pricing offers for on-demand shared EC2 c5.2xlarge without pre-installed software, with Linux OS
+
+```sql
+select
+  *
```
cbruno10, line 11: Instead of `*`, can you please select some key columns to show? We generally don't like using `*` because when there are a lot of columns it can be overwhelming (and there are quite a lot in this table), and it also doesn't give users an initial direction on what info to check out/what's important.

docs/tables/aws_pricing_product.md
```diff
@@ -0,0 +1,55 @@
+# Table: aws_pricing_product
+
+Returns costs, and offer definition, for the corresponding service.
+
+## Examples
```
cbruno10, line 5: @aminvielledebatAtBedrock In the PR body (or in a new comment in it), can you please post the results (or a screenshot if it's too long), so we can see what some rows look like?

aws/table_aws_pricing_product.go
```diff
@@ -0,0 +1,274 @@
...
+			plugin.Logger(ctx).Error("aws_pricing_product.listPricingProduct", "api_error", err)
+			return nil, err
+		}
+
+		for _, priceList := range output.PriceList {
+			var priceListObject PriceList
+
+			err = json.Unmarshal([]byte(priceList), &priceListObject)
+			if err != nil {
+				plugin.Logger(ctx).Error("listPricingProduct unmarshal", "err", err)
```
cbruno10, line 198:
```suggestion
				plugin.Logger(ctx).Error("aws_pricing_product.listPricingProduct", "unmarshal_err", err)
```

aws/table_aws_pricing_product.go
cbruno10, line 45:
```suggestion
			{Name: "price_per_unit", Description: "Price by unit", Type: proto.ColumnType_DOUBLE, Transform: transform.FromField("Offer.PriceDimension.PricePerUnit").Transform(extractPrice)},
```

I think I'd keep the column name the same as the field name, price per unit is more descriptive and closer to what it represents, whereas just price is more ambiguous.

aws/table_aws_pricing_product.go
```diff
@@ -0,0 +1,274 @@
...
+			{Name: "tenancy", Description: "The type of tenancy allowed on the Amazon EC2 instance.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.tenancy")},
+			{Name: "license_model", Description: "Describes the license model for your instance.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.licenseModel")},
+			{Name: "current_generation", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.currentGeneration")},
+			{Name: "pre_installed_software", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.preInstalledSw")},
+			{Name: "processor_architecture", Description: "Describes your processor architecture.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.processorArchitecture")},
+			// multi az attributes
+			{Name: "deployment_option", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.deploymentOption")},
+			// RDS fields
+			{Name: "database_engine", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.databaseEngine")},
```
cbruno10, line 73: Are there other services beside RDS and ElastiCache that have their own fields? If so, do you have a reference list you're using? I'm concerned about scaling this table with specific columns if there are a lot of services/resources with their own specific properties, as this could result in a lot of columns returning `nil` since they'd never have those attributes.

aws/table_aws_pricing_product.go
```diff
@@ -0,0 +1,274 @@
...
+			{Name: "min_volume_size", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.minVolumeSize")},
+			{Name: "maxVolumeSize", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.maxVolumeSize")},
+			// Elasticache fields
+			{Name: "cache_engine", Description: "", Type: proto.ColumnType_STRING, Transform: transform.FromField("Product.Attributes.cacheEngine")},
+			// all attributes
+			{Name: "term_attributes", Type: proto.ColumnType_JSON, Transform: transform.FromField("Product.Attributes")},
```
cbruno10, line 79: If we're parsing out the attributes above, I'm not sure we should include this column along with them. It feels like duplicate info, and I imagine this field could be pretty large. Is there a particular use case you thought of when including this?

cbruno10 (changes requested):
@aminvielledebatAtBedrock Thanks again for a new PR! I've included some questions and suggestions as well, can you please have a look? If you have any questions, please let me know.

aws/table_aws_pricing_product.go
aminvielledebatAtBedrock: Hi @cbruno10 and thanks again for your high qualified review :-)

See bellow an example of price get back from the API : 

``` json
"VJKEFXHAX2JKR9HH.JRTCKXETXF.6YS6EN2CT7": {
            "unit": "Hrs",
            "endRange": "Inf",
            "description": "$0.404 per Unused Reservation Linux c5.2xlarge Instance Hour",
            "appliesTo": [],
            "rateCode": "VJKEFXHAX2JKR9HH.JRTCKXETXF.6YS6EN2CT7",
            "beginRange": "0",
            "pricePerUnit": {
              "USD": "0.4040000000"
            }
```

So, pricePerUnit is a map, with only one key:value row.
This is why i return the first record in my for loop : 

``` go
func extractCurrency(_ context.Context, d *transform.TransformData) (interface{}, error) {
	priceMap := d.Value.(map[string]*string)

	for currency := range priceMap {
		return currency, nil
	}
	return nil, nil
}
```

Feel free to ask me any change if you have a better idea

aws/table_aws_pricing_product.go
aminvielledebatAtBedrock: Well spotted ! I missed it !

docs/tables/aws_pricing_product.md
aminvielledebatAtBedrock: Done in the PR description.

Feel free to ask me any other examples

aws/table_aws_pricing_product.go
aminvielledebatAtBedrock: You're right ! There are other columns wich are not described in my PR.
I was focused on the most used one's but i think it could be interesting to have all available attributes.

For that, i can use the output of this request : 
``` sql
select
    attribute_name
    ,count(*) as nb
from bedrock_root_us.aws_pricing_service_attribute
group by 1
```

| attribute\_name            | usage |
| -------------------------- | ----- |
| analyticsMode              | 12    |
| awsresource                | 11    |
| baseProductReferenceCode   | 9     |
| bestPractices              | 63    |
| brokerEngine               | 21    |
| bundle                     | 9     |
| cacheType                  | 17    |
| calltype                   | 28    |
| category                   | 10    |
| ciType                     | 11    |
| clientLocation             | 13    |
| computeType                | 4     |
| computeprovider            | 16    |
| deploymentModelDescription | 10    |
| description                | 5     |
| device                     | 10    |
| executionLocation          | 11    |
| extraDayFee                | 27    |
| findingGroup               | 13    |
| flow                       | 10    |
| flowType                   | 18    |
| fromLocation               | 13    |
| fromcountry                | 11    |
| group                      | 645   |
| groupDescription           | 5     |
| integratingApi             | 13    |
| location                   | 28    |
| lookoutvisionimage         | 10    |
| mailboxStorage             | 8     |
| meterMode                  | 8     |
| minimumStorageVolume       | 13    |
| ondemand                   | 11    |
| operation                  | 890   |
| operationType              | 10    |
| opsItems                   | 24    |
| processorFeatures          | 25    |
| productType                | 11    |
| querytype                  | 14    |
| regionCode                 | 5     |
| resourceAssessment         | 10    |
| resourceType               | 95    |
| scanType                   | 12    |
| servicecode                | 8     |
| servicename                | 383   |
| storageClass               | 6     |
| tenancySupport             | 1     |
| termType                   | 6     |
| transactionType            | 1     |
| type                       | 9     |
| usagetype                  | 23    |
| vcpu                       | 2     |
| videoCodec                 | 16    |
| videoResolution            | 8     |
| vmwareregion               | 17    |
| volumeType                 | 1     |
| withActiveUsers            | 9     |

If you're ok with that, I can do the change quickly

aws/table_aws_pricing_product.go
aminvielledebatAtBedrock: You're right : for rds, ec2, and elasticache instances, it's duplicate info.

My goal was to allow users to fetch their custom attributes for a specific service_code which is not already supported by the table.

I think adding others columns is better, even if there will be a lot of nil values

aws/table_aws_pricing_product.go
cbruno10: Does it always just return `USD`, or does it sometimes return the local currency, e.g., `CAD` as a key? Do you know if there are any cases where it could return multiple price per unit currencies?

aws/table_aws_pricing_product.go
aminvielledebatAtBedrock: Ok i was on a good way. It seems the `aws_pricing_service_attribute` has a bug with a pointer.

Some column seems to be duplicated whereas the json returned by aws is valid.

I will take a look on that, and maybe use the exported attributes for my table. It seems to be a lot of them :pleading_face:

aminvielledebatAtBedrock: Fixed in https://github.com/turbot/steampipe-plugin-aws/pull/1372

aminvielledebatAtBedrock: The exhausted list 

| attribute\_name                    | nb  |
| ---------------------------------- | --- |
| 1yearPrice                         | 1   |
| 3yearPrice                         | 1   |
| LeaseContractLength                | 6   |
| OfferingClass                      | 6   |
| PurchaseOption                     | 6   |
| Restriction                        | 48  |
| acceleratorSize                    | 1   |
| acceleratorType                    | 1   |
| accessPointType                    | 1   |
| accessType                         | 1   |
| accountAssistance                  | 3   |
| action                             | 1   |
| activityType                       | 1   |
| acu                                | 1   |
| addonFeature                       | 1   |
| alarmType                          | 1   |
| analyticsMode                      | 1   |
| analyticsType                      | 1   |
| apiCalls                           | 1   |
| apiType                            | 1   |
| apirequestsfree                    | 1   |
| apirequestspaid                    | 1   |
| architecturalReview                | 3   |
| architectureSupport                | 3   |
| attachmentType                     | 2   |
| automaticLabel                     | 1   |
| availability                       | 4   |
| availabilityZone                   | 1   |
| availabilityzone                   | 1   |
| awsiotevents                       | 1   |
| awsresource                        | 1   |
| backupEvaluations                  | 1   |
| backupservice                      | 1   |
| bandwidth                          | 1   |
| baseProductReferenceCode           | 1   |
| bestPractices                      | 3   |
| bitRate                            | 1   |
| brioproductid                      | 1   |
| brokerEngine                       | 1   |
| build                              | 1   |
| bundle                             | 3   |
| bundleDescription                  | 1   |
| bundleGroup                        | 1   |
| cacheEngine                        | 1   |
| cacheMemorySizeGb                  | 1   |
| cacheType                          | 1   |
| cachememorysize                    | 1   |
| callingType                        | 3   |
| calltype                           | 2   |
| capability                         | 1   |
| capacity                           | 1   |
| capacitystatus                     | 2   |
| caseSeverityresponseTimes          | 3   |
| category                           | 2   |
| changerequestsfree                 | 1   |
| changerequestspaid                 | 1   |
| channel                            | 1   |
| chargeid                           | 1   |
| ciType                             | 1   |
| classicnetworkingsupport           | 1   |
| classificationType                 | 1   |
| clientLocation                     | 1   |
| clockSpeed                         | 8   |
| cloudSearchVersion                 | 1   |
| cloudformationresourceProvider     | 1   |
| clustertype                        | 1   |
| comments                           | 1   |
| commitmentDays                     | 1   |
| component                          | 1   |
| compute                            | 1   |
| computeFamily                      | 2   |
| computeType                        | 3   |
| computeprovider                    | 1   |
| computesize                        | 1   |
| computetype                        | 1   |
| concurrencyscalingfreeusage        | 1   |
| connectionType                     | 1   |
| contentSource                      | 1   |
| contentType                        | 2   |
| country                            | 5   |
| countsAgainstQuota                 | 1   |
| cpuArchitecture                    | 1   |
| cputupe                            | 1   |
| cputype                            | 2   |
| currentGeneration                  | 13  |
| customerServiceAndCommunities      | 3   |
| data                               | 1   |
| dataAction                         | 1   |
| dataTransfer                       | 1   |
| dataTransferQuota                  | 1   |
| dataType                           | 1   |
| databaseEdition                    | 1   |
| databaseEngine                     | 3   |
| datastoreStoragetype               | 1   |
| datatransferout                    | 1   |
| datatype                           | 1   |
| dedicatedEbsThroughput             | 5   |
| deploymentLocation                 | 1   |
| deploymentModel                    | 3   |
| deploymentModelDescription         | 1   |
| deploymentOption                   | 4   |
| describes                          | 1   |
| description                        | 26  |
| device                             | 1   |
| deviceDescription                  | 1   |
| deviceName                         | 1   |
| deviceOs                           | 1   |
| deviceProvider                     | 1   |
| deviceType                         | 1   |
| devicename                         | 1   |
| dimension                          | 1   |
| directConnectLocation              | 1   |
| directorySize                      | 1   |
| directoryType                      | 1   |
| directoryTypeDescription           | 1   |
| disableactivationconfirmationemail | 1   |
| dominantnondominant                | 1   |
| durability                         | 3   |
| ebsOptimized                       | 1   |
| ecu                                | 5   |
| edition                            | 1   |
| elasticGraphicsType                | 1   |
| endpoint                           | 1   |
| endpointType                       | 2   |
| engine                             | 2   |
| engineCode                         | 3   |
| engineMediaType                    | 1   |
| enhancedNetworkingSupport          | 1   |
| enhancedNetworkingSupported        | 6   |
| entityType                         | 1   |
| eventType                          | 4   |
| executionFrequency                 | 1   |
| executionLocation                  | 1   |
| executionMode                      | 1   |
| externalInstanceType               | 1   |
| extraDayFee                        | 1   |
| feature                            | 1   |
| feeCode                            | 6   |
| feeDescription                     | 6   |
| feecode                            | 1   |
| fileCacheType                      | 1   |
| fileSystemType                     | 1   |
| findingGroup                       | 1   |
| findingSource                      | 1   |
| findingStorage                     | 1   |
| flow                               | 1   |
| flowType                           | 1   |
| frameRate                          | 1   |
| frameworkType                      | 1   |
| freeDays                           | 1   |
| freeOverage                        | 1   |
| freeQueryTypes                     | 1   |
| freeTier                           | 1   |
| freeTrial                          | 1   |
| freeUsageIncluded                  | 2   |
| freetier                           | 1   |
| frequency                          | 1   |
| frequencyMode                      | 1   |
| fromLocation                       | 12  |
| fromLocationType                   | 12  |
| fromRegionCode                     | 10  |
| fromcountry                        | 3   |
| gb                                 | 1   |
| georegioncode                      | 1   |
| gets                               | 1   |
| gpu                                | 3   |
| gpuMemory                          | 2   |
| granularity                        | 1   |
| graphqloperation                   | 1   |
| groundstation                      | 1   |
| group                              | 66  |
| groupDescription                   | 52  |
| groupDetails                       | 1   |
| highAvailability                   | 1   |
| includedEvents                     | 1   |
| includedServices                   | 3   |
| indexingSource                     | 1   |
| ingestType                         | 1   |
| input                              | 1   |
| inputMode                          | 2   |
| inputType                          | 1   |
| inputtype                          | 1   |
| insightstype                       | 1   |
| instance                           | 1   |
| instanceCapacity10xlarge           | 1   |
| instanceCapacity12xlarge           | 1   |
| instanceCapacity16xlarge           | 2   |
| instanceCapacity18xlarge           | 1   |
| instanceCapacity24xlarge           | 1   |
| instanceCapacity2xlarge            | 2   |
| instanceCapacity32xlarge           | 1   |
| instanceCapacity4xlarge            | 2   |
| instanceCapacity8xlarge            | 2   |
| instanceCapacity9xlarge            | 1   |
| instanceCapacityLarge              | 1   |
| instanceCapacityMedium             | 1   |
| instanceCapacityMetal              | 1   |
| instanceCapacityXlarge             | 2   |
| instanceFamily                     | 13  |
| instanceFunction                   | 1   |
| instanceName                       | 1   |
| instanceStorageGb                  | 1   |
| instanceType                       | 21  |
| instanceTypeFamily                 | 2   |
| instances                          | 1   |
| instancesku                        | 1   |
| instancetype                       | 1   |
| integratingApi                     | 2   |
| integratingFeature                 | 2   |
| integratingService                 | 2   |
| intelAvx2Available                 | 2   |
| intelAvxAvailable                  | 2   |
| intelTurboAvailable                | 2   |
| io                                 | 1   |
| ioRequestType                      | 1   |
| isSpotInstance                     | 1   |
| iscommitcpsku                      | 1   |
| isfree                             | 1   |
| isshadow                           | 1   |
| iswebsocket                        | 1   |
| jobFee                             | 1   |
| jobnshipp                          | 1   |
| labelingTaskType                   | 1   |
| launchSupport                      | 3   |
| learningMode                       | 1   |
| license                            | 1   |
| licenseModel                       | 6   |
| licenseType                        | 3   |
| lineType                           | 1   |
| location                           | 197 |
| locationType                       | 197 |
| lockeprofiles                      | 1   |
| logsDestination                    | 1   |
| logsSource                         | 1   |
| logsType                           | 1   |
| lookoutvisionimage                 | 1   |
| machineLearningProcess             | 1   |
| mailboxStorage                     | 1   |
| marketoption                       | 1   |
| maxIopsBurstPerformance            | 1   |
| maxIopsvolume                      | 1   |
| maxThroughputvolume                | 2   |
| maxVolumeSize                      | 5   |
| maximumCapacity                    | 2   |
| maximumExtendedStorage             | 1   |
| maximumStorageVolume               | 1   |
| memory                             | 17  |
| memoryGib                          | 4   |
| memorygb                           | 1   |
| memorytype                         | 2   |
| messageDeliveryFrequency           | 1   |
| messageDeliveryOrder               | 1   |
| messageType                        | 1   |
| messagetype                        | 2   |
| meterMode                          | 1   |
| meteringType                       | 1   |
| meterunit                          | 1   |
| minVolumeSize                      | 4   |
| minimumStorageVolume               | 1   |
| networkPerformance                 | 11  |
| newcode                            | 2   |
| nlpclass                           | 1   |
| normalizationSizeFactor            | 6   |
| numbertype                         | 3   |
| offer                              | 1   |
| ondemand                           | 1   |
| operatingSystem                    | 9   |
| operation                          | 105 |
| operationOutcome                   | 1   |
| operationType                      | 1   |
| operationsSupport                  | 3   |
| opsItems                           | 1   |
| origin                             | 1   |
| os                                 | 1   |
| osLicenseModel                     | 1   |
| outpostusage                       | 1   |
| output                             | 1   |
| outputMode                         | 2   |
| outputType                         | 1   |
| overageType                        | 1   |
| overhead                           | 1   |
| parameterType                      | 1   |
| physicalCores                      | 1   |
| physicalCpu                        | 1   |
| physicalGpu                        | 1   |
| physicalProcessor                  | 6   |
| pipeline                           | 2   |
| planType                           | 1   |
| platofeaturetype                   | 1   |
| platoinstancename                  | 1   |
| platopagedatatype                  | 1   |
| platoprotectionpolicytype          | 1   |
| platousagetype                     | 1   |
| portSpeed                          | 1   |
| power                              | 1   |
| preInstalledSw                     | 2   |
| predictionType                     | 1   |
| pricingUnit                        | 9   |
| pricingplan                        | 1   |
| primaryplaceofuse                  | 4   |
| proactiveGuidance                  | 3   |
| processorArchitecture              | 7   |
| processorFeatures                  | 6   |
| productFamily                      | 193 |
| productName                        | 1   |
| productSchemaDescription           | 1   |
| productType                        | 2   |
| productgroupid                     | 1   |
| productsubgroup                    | 1   |
| producttype                        | 1   |
| programmaticCaseManagement         | 3   |
| protocol                           | 2   |
| provider                           | 2   |
| provisioned                        | 2   |
| querytype                          | 1   |
| queueType                          | 1   |
| readtype                           | 1   |
| realtimeoperation                  | 1   |
| recipient                          | 1   |
| regionCode                         | 158 |
| releaseType                        | 1   |
| replicationType                    | 2   |
| requestDescription                 | 1   |
| requestType                        | 3   |
| reserveType                        | 1   |
| reservedTier                       | 1   |
| resolution                         | 1   |
| resource                           | 1   |
| resourceAssessment                 | 1   |
| resourceEndpoint                   | 2   |
| resourcePriceGroup                 | 1   |
| resourceType                       | 3   |
| response                           | 1   |
| rootvolume                         | 1   |
| routingTarget                      | 2   |
| routingType                        | 2   |
| runningMode                        | 1   |
| scanType                           | 1   |
| serverLocation                     | 1   |
| servicecode                        | 199 |
| servicename                        | 186 |
| simulatorName                      | 1   |
| sitelinkOriginPopGroup             | 1   |
| size                               | 1   |
| snapshotarchivefeetype             | 1   |
| snowballType                       | 2   |
| softwareIncluded                   | 1   |
| softwareType                       | 1   |
| sourcetype                         | 1   |
| ssd                                | 1   |
| standardGroup                      | 1   |
| standardStorage                    | 1   |
| standardStorageRetentionIncluded   | 1   |
| steps                              | 1   |
| storage                            | 13  |
| storageCapacityQuota               | 1   |
| storageClass                       | 8   |
| storageDescription                 | 2   |
| storageFamily                      | 1   |
| storageMedia                       | 10  |
| storageType                        | 4   |
| storageclass                       | 2   |
| storagetype                        | 4   |
| subcategory                        | 1   |
| subscriptionType                   | 1   |
| subservice                         | 1   |
| supportedModes                     | 2   |
| technicalSupport                   | 3   |
| tenancy                            | 4   |
| tenancySupport                     | 1   |
| termType                           | 199 |
| thirdpartySoftwareSupport          | 3   |
| throughput                         | 1   |
| throughputCapacity                 | 1   |
| throughputClass                    | 1   |
| tickettype                         | 1   |
| tier                               | 4   |
| tiertype                           | 1   |
| toLocation                         | 12  |
| toLocationType                     | 13  |
| toRegionCode                       | 9   |
| tocountry                          | 5   |
| trafficDirection                   | 1   |
| training                           | 3   |
| transactionType                    | 3   |
| transactiontype                    | 1   |
| transcodingResult                  | 2   |
| transferType                       | 12  |
| trialProduct                       | 1   |
| type                               | 8   |
| updates                            | 1   |
| upfrontCommitment                  | 1   |
| usageFamily                        | 1   |
| usageTier                          | 1   |
| usageVolume                        | 1   |
| usagetype                          | 199 |
| usertype                           | 1   |
| uservolume                         | 1   |
| vCpu                               | 1   |
| vcpu                               | 20  |
| version                            | 5   |
| videoCodec                         | 1   |
| videoFrameRate                     | 1   |
| videoMemoryGib                     | 1   |
| videoQuality                       | 1   |
| videoQualitySetting                | 1   |
| videoResolution                    | 2   |
| virtualInterfaceType               | 1   |
| vmwareproductid                    | 1   |
| vmwareregion                       | 1   |
| volumeApiName                      | 1   |
| volumeName                         | 1   |
| volumeType                         | 16  |
| vpcnetworkingsupport               | 1   |
| whoCanOpenCases                    | 3   |
| withActiveUsers                    | 1   |
| workforceType                      | 2   |

cbruno10: @aminvielledebatAtBedrock Do you also know what the breakout of these attributes are per service? My concern is that if there are a large number of services, and some attributes are only present in a few services each, that if a user runs `select * from aws_pricing_product where service_code = '...'`, they may get back a lot of columns with `nil` data, which is a lot to sort through, and may cause confusion, i.e., is this attribute expected to have data and the table didn't return it correctly?

For instance, `volume_name` will only be in the EC2/EBS product most likely, but will be `nil` for all others. With the updated listed, it seems like there are a lot of attributes that only appear once, so there will be a lot of `nil` column values coming through consistently.

Not sure if this is the right direction, but I wonder what if we broke up the `aws_pricing_product` table into multiple tables, one per service code? It'd be a lot of tables, but would that help with reducing the number of `nil` attributes?

Alternatively, we can avoid adding any specific attributes as columns and rely on users to pull them out, e.g., `select attributes -> 'volume_name' as volume_name from aws_pricing_product where service_code = '...'`, while leaving the common attributes, like `usage_type` in as columns.

aminvielledebatAtBedrock: I don't know what's the better option, if we take account the following points : 
- Only 9 attributes (on 417) are used by the most part of aws services (for 199 referenced services)
  There are too common, and don't allow the end user to filter the results
  For example, AmazonElasticCache on `eu-west-3` returns 861 rows :disappointed: 
  The only difference between rows will be the json attributes columns. Not so good IMO.
- Attributes are used by the API for filtering. Not using them involves fetching all data per `service_code` 
- 417 attributes is too much for one table. I agree with you.
- We are talking about 199 different services. I really like steampipe, but not enough to write 199 tables :laughing: 

Do you think dynamic tables could be possible (with a first call to `DescribeServices` to fetch column definitions) ? 

Or the least bad could be keeping only 9 attributes, with an json attributes column... 

I'll follow you !

aws/table_aws_pricing_product.go
aminvielledebatAtBedrock: We have always paid aws with `USD` :laughing:

More seriously, you can choose you currency in your aws account, but it seems everything is not converted magicaly.
See [here](https://aws.amazon.com/premiumsupport/knowledge-center/non-us-currency/?nc1=h_ls)

> When using a non-USD currency, consider the following:
> All preferred currency amounts shown in the AWS Billing and Cost Management console are estimated in USD.

So, it's not planned to have several currencies in pricing API apparently. I can hardcode `USD` key in the map if you want.

cbruno10: I think it's ok to pull the first currency from the map if we only expect one item to be returned in the map.

aws/table_aws_pricing_product.go
cbruno10: @aminvielledebatAtBedrock Thanks for the figures, they're really helpful in understanding what the API returns and how the data looks across the services. Let me have a look through the data and service codes again, and I'll respond here again tomorrow with any findings/suggestions I have.

Overall, my thoughts are:
- One table seems like the best design for now. We could create 199+ tables dynamically, but I'm not sure I'd want to flood the schema with all of those tables.
- With 417 attributes, that's a lot to try map manually into columns. I'm not sure if there's a way to map automatically without making the API calls (maybe there's a reference in the AWS SDK code though, like there are endpoints).
- We could still implement filtering by attribute by adding a `filters` JSON column that serves as a `ListConfig` key column. For instance, taking one example from your current table, it would then look something like:
```
select
  service_code,
  attributes -> 'instanceType' as instance_type,
  attributes -> 'instanceName' as instance_name,
  price_per_unit::numeric::money
from
  aws_pricing_product 
where
  service_code = 'AmazonEC2' 
  and region_code = 'eu-west-3' 
  and location_type = 'AWS Region' 
  and filters = '{
    "instanceType": "c5.2xlarge",
    "operatingSystem": "Linux",
    "tenancy": "Shared",
    "preInstalledSw": "NA"
    "capacityStatus": "Used"
  }'::jsonb;
  ```
  - The `service_code`, `region_code`, and `location_type` columns would be actual columns (based on the counts above), while the others would be passed through the new `filters` column. If key columns are passed in along with `filter`, like in the example above, they should all be combined into a single `input.Filters`.
  - Note: I actually prefer using `jsonb_build_object` instead, but there's currently a bug in the Steampipe SDK preventing that from working correctly with JSON type key columns.
  
  I believe the design above would give users the ability to query pricing products, top column support for common columns, and the flexibility to pass in other attributes for filtering.
  
  @aminvielledebatAtBedrock @johnsmyth - Curious to hear your thoughts and feedback, thanks!

johnsmyth: @cbruno10 I think that your approach seems reasonable, but it will be pretty hard for people to use.  I like @aminvielledebatAtBedrock 's idea of dynamic tables, but we would end up with alot of dynamic tables.   So maybe we can do both:
- add the 'generic' table to the plugin as @cbruno10 suggested
- add the per-service tables (dynamic tables if that works, else manually curated) but behind a config argument:
    - These tables would NOT be enabled by default
    - The argument would ideally accept an array of table (service) names, and allow the '*' wildcard.  (similar to the how the `regions` arg works)
        - This would allows users to only enable the tables they want
        - This would also allow user to enable it in only a single connection, which is probably desirable when using an aggregator (otherwise you would get the same pricing info from every connection in the aggregator).
        
thoughts?

cbruno10: @johnsmyth I like the idea of putting the dynamic tables behind a config argument, and agree with the other design notes in your comment.

For the scope of this PR, I think we could start with the generic table, and then create an issue to track adding dynamic tables, as that would be considerable more work to add, e.g., decide on table and column naming structure that works for all service/service codes, test API limits in case someone had `*`, meaning they want all tables (each of which would require a separate API call I believe, as you can only pass 1 service code per request), test the time it takes to create the tables.

aminvielledebatAtBedrock: Done. Could you have a look @cbruno10 please ?

cbruno10:
@aminvielledebatAtBedrock This is on my radar to review again tomorrow. Also, please note that you don't have to force push each time if you're trying to keep your commit history tidy, we squash as part of merging.

cbruno10 (approved):

cbruno10:
Thanks again @aminvielledebatAtBedrock for all of the hard work and working through multiple review cycles!

PR #1376, opened 2022-11-04
Remove v2 migration column compatible changes and fix other duplicate data issues

LalitTurbot:


aws/errors.go
```diff
@@ -5,29 +5,19 @@ import (
...
 
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/smithy-go"
 	"github.com/turbot/steampipe-plugin-sdk/v4/plugin"
 )
 
-// isNotFoundError:: function which returns an ErrorPredicate for AWS API calls
-func isNotFoundError(notFoundErrors []string) plugin.ErrorPredicateWithContext {
+// isNotFoundErrorV2:: function which returns an ErrorPredicate for AWS API calls
+func isNotFoundErrorV2(notFoundErrors []string) plugin.ErrorPredicateWithContext {
```
cbruno10, line 13: Do we still need to call this `V2` if there isn't a v1 version anymore?

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/turbot/steampipe-plugin-aws
 go 1.19
 
 require (
-	github.com/aws/aws-sdk-go v1.44.49
+	github.com/aws/aws-sdk-go v1.44.130
```
cbruno10, line 6: Why are we bumping this version? Is it for support of new endpoints?

cbruno10 (changes requested):
@LalitTurbot Please see comments, thanks!

go.mod
LalitTurbot: > Why are we bumping this version? Is it for support of new endpoints?

Yes, for new endpoints

aws/errors.go
```diff
@@ -5,29 +5,19 @@ import (
...
 	return func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData, err error) bool {
 		awsConfig := GetConfig(d.Connection)
 
 		// If the get or list hydrate functions have an overriding IgnoreConfig
 		// defined using the isNotFoundError function, then it should
 		// also check for errors in the "ignore_error_codes" config argument
 		allErrors := append(notFoundErrors, awsConfig.IgnoreErrorCodes...)
-		if awsErr, ok := err.(awserr.Error); ok {
```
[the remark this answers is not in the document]
LalitTurbot: > Do we handle these elsewhere? Same for the removal of code below.

Not now
That was the way to handle to v1 errors

aws/errors.go
LalitTurbot: > Do we still need to call this `V2` if there isn't a v1 version anymore?

No required, I left it because it will lead to changes to so many files.

It can be just find and replace in a single step if we want to do it.

cbruno10 (approved):

PR #1437, opened 2022-11-29
Add table aws_cloudtrail_query closes #1314

### LINKED ISSUE #1314: Add table aws_cloudtrail_lake_query
Ability to query CloudTrail lake queries with Steampipe.  Provides an alternative to query and join through Steampipe, and alternative option to the cloudtrail event table through cloudwatch log groups already supported

https://awscli.amazonaws.com/v2/documentation/api/latest/reference/cloudtrail/start-query.html

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
N/A
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
> select
  query_id,
  event_data_store_arn,
  query_status,
  query_status,
  creation_time,
  events_matched,
  events_scanned
from
  aws_cloudtrail_lake_query;
+--------------------------------------+-----------------------------------------------------------------------------------------------+--------------+--------------+---------------------------+--------
| query_id                             | event_data_store_arn                                                                          | query_status | query_status | creation_time             | events_
+--------------------------------------+-----------------------------------------------------------------------------------------------+--------------+--------------+---------------------------+--------
| 57f1ad01-9f12-416a-bd24-7e233ac60515 | arn:aws:cloudtrail:us-east-1:457402157489:eventdatastore/03543d1c-054c-4016-95ae-1f8faef24485 | FINISHED     | FINISHED     | 2022-11-29T11:43:51+05:30 | 0      
| 64ca2d09-0ce2-403b-97c3-57bbd186e0ea | arn:aws:cloudtrail:us-east-1:457402157489:eventdatastore/5c5ae735-06f1-40b3-84d5-d9838162b1d1 | FINISHED     | FINISHED     | 2022-11-24T17:36:02+05:30 | 0      
+--------------------------------------+-----------------------------------------------------------------------------------------------+--------------+--------------+---------------------------+--------

> select
  q.query_id as query_id,
  q.event_data_store_arn as event_data_store_arn,
  s.name as event_data_store_name,
  s.status as event_data_store_status,
  s.multi_region_enabled as multi_region_enabled,
  s.termination_protection_enabled as termination_protection_enabled,
  s.updated_timestamp as event_data_store_updated_timestamp
from
  aws_cloudtrail_lake_query as q,
  aws_cloudtrail_event_data_store as s
where
 s.event_data_store_arn = q.event_data_store_arn;

+--------------------------------------+-----------------------------------------------------------------------------------------------+-----------------------+-------------------------+----------------
| query_id                             | event_data_store_arn                                                                          | event_data_store_name | event_data_store_status | multi_region_en
+--------------------------------------+-----------------------------------------------------------------------------------------------+-----------------------+-------------------------+----------------
| 64ca2d09-0ce2-403b-97c3-57bbd186e0ea | arn:aws:cloudtrail:us-east-1:457402157489:eventdatastore/5c5ae735-06f1-40b3-84d5-d9838162b1d1 | test-event-data-store | ENABLED                 | true           
| 57f1ad01-9f12-416a-bd24-7e233ac60515 | arn:aws:cloudtrail:us-east-1:457402157489:eventdatastore/03543d1c-054c-4016-95ae-1f8faef24485 | turbottest51242       | ENABLED                 | true           
+--------------------------------------+-----------------------------------------------------------------------------------------------+-----------------------+-------------------------+----------------

```
</details>


aws/table_aws_cloudtrail_lake_query.go
```diff
@@ -0,0 +1,199 @@
...
+		if limit < maxLimit {
+			if limit < 1 {
+				maxLimit = 1
+			} else {
+				maxLimit = limit
+			}
+		}
+	}
+
+	input := &cloudtrail.ListQueriesInput{
```
cbruno10, line 144: From https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudtrail#Client.ListQueries, it seems like only the queries from the past 7 days are returned. How would we expect users to know this and how would we expect them to interact with the table? Do they need to specify `creation_time` in a query then?

aws/table_aws_cloudtrail_lake_query.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/turbot/steampipe-plugin-sdk/v4/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v4/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v4/plugin/transform"
+)
+
+//// TABLE DEFINITION
+
+func tableAwsCloudtrailLakeQuery(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_cloudtrail_lake_query",
```
cbruno10, line 20: Should it be `aws_cloudtrail_query` instead? Looking at the API, there's no mention of `Lake` (https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_ListQueries.html), though other AWS docs do associate them with CloudTrail Lake (https://docs.aws.amazon.com/awscloudtrail/latest/userguide/query-create-edit-query.html).

aws/table_aws_cloudtrail_lake_query.go
```diff
@@ -0,0 +1,199 @@
...
+		Description: "AWS CloudTrail Lake Query",
+		List: &plugin.ListConfig{
+			ParentHydrate: listCloudTrailEventDataStores,
```
cbruno10, line 23: Do we want a parent hydrate here, or should a user need to enter an event data store ARN? How many event data stores can be in a region and how many queries can an event data store have? What's the API performance like if there are a large number of both?

aws/table_aws_cloudtrail_lake_query.go
```diff
@@ -0,0 +1,199 @@
...
+				{
+					Name:    "event_data_store_arn",
+					Require: plugin.Optional,
+				},
+			},
+		},
+		GetMatrixItemFunc: BuildRegionList,
+		Columns: awsRegionalColumns([]*plugin.Column{
+			{
+				Name:        "query_id",
```
cbruno10, line 35: We should support `query_id` as an optional key column

cbruno10 (changes requested):
@ParthaI Please see comments, thanks!

aws/table_aws_cloudtrail_lake_query.go
ParthaI: We can create five event source per region https://docs.aws.amazon.com/awscloudtrail/latest/userguide/WhatIsCloudTrail-Limits.html

aws/table_aws_cloudtrail_lake_query.go
ParthaI: The list call does not support `QueryId` as an input param

aws/table_aws_cloudtrail_lake_query.go
ParthaI: There is no data available in our console, which is older than seven days, so we could not test this scenario.
But if we pass the `creation_time` value in the where clause, it returns the result instead.

aws/table_aws_cloudtrail_lake_query.go
ParthaI: Updated the table name.

aws/table_aws_cloudtrail_lake_query.go
cbruno10: Where did you find that figure? On that page, I can see:
```
Resource | Default Limit | Comments
Trails per region | 5 | This limit cannot be increased.
Event data stores in CloudTrail Lake | 10 | This limit cannot be increased.
```

So does that mean 10 per trail, region, or account? Or is it 5 instead of 10?

aws/table_aws_cloudtrail_lake_query.go
cbruno10: I believe the table doc should mention the default behaviour (the table returns queries from the past 7 days), and we should ask users to specify a creation date if they want results from further back. We should also add an example showing how to specify `creation_time`.

aws/table_aws_cloudtrail_lake_query.go
ParthaI: Sorry it is ten

aws/table_aws_cloudtrail_lake_query.go
ParthaI: @cbruno10 The `DescribeQuery` API doesn't provide all the properties we are getting from list call, so we can not use it in `GetConfig`. We can use this API as a hydrate call to fetch some other information that is expressly provided by this API.

aws/table_aws_cloudtrail_lake_query.go
cbruno10: Agree, using the `StartTime` and `EndTime` attributes is the right direction. How many more days do we need to wait from today, December 6, in order to have queries older than 7 days?

aws/table_aws_cloudtrail_lake_query.go
cbruno10: Sometimes they set limits high, even if the number of resources allowed per region or account is very small.

I'm still unclear on how many event data stores are allowed in CloudTrail overall, as per my previous question. Can you please find out if it's 10 per trail, region, or account?

docs/tables/aws_cloudtrail_query.md
```diff
@@ -0,0 +1,74 @@
...
+  s.termination_protection_enabled as termination_protection_enabled,
+  s.updated_timestamp as event_data_store_updated_timestamp
+from
+  aws_cloudtrail_query as q,
+  aws_cloudtrail_event_data_store as s
+where
+ s.event_data_store_arn = q.event_data_store_arn;
+```
+
+## List queries created within the last 30 days
```
cbruno10, line 60:
```suggestion
## List queries created within the last 3 days
```

From our other discussion, since we're assuming that queries from beyond 7 days ago cannot be returned, we should update this example to restrict further within that 7 day range.

aws/table_aws_cloudtrail_lake_query.go
ParthaI: @cbruno10 We can create upto 10 Event Data Store per trail, I tried to create more than 10 Event Data Store but got the limit exceeded error as below.

```
EventDataStoreMaxLimitExceededException
The maximum number of event data stores, 10, has already been created. This limit applies to the total number of event data stores that you can create in all regions.
```

aws/table_aws_cloudtrail_query.go
```diff
@@ -0,0 +1,242 @@
...
+			},
+		}),
+	}
+}
+
+// We need custom structure
+// 1. EventDataStoreArn is required for joing the queries
+// 2. list/get call do not provide the EventDataStoreArn
+// 3. Both list/get API call return different stracture with dfferent data
+type QueryInfo struct {
```
cbruno10, line 122: @ParthaI Instead of having just 1 custom structure, could we have 2 structures instead, 1 for List results and 1 for Get results? Each of these would include the event data store ARN and then also contain the AWS SDK structure related to the List or Get call (so we don't have to redefine each property).

I think this would allow us to lean into the structures AWS has defined while adding any properties, like the event data store as well.

Another advantage would be it would be less likely to miss columns, like we are today, as the DescribeQuery response has attributes we don't have as columns today. So any missing columns, like the delivery status, delivery S3 URI, and error message should be added.

Our columns would then use Transform.FromFields and leverage the ability to specify multiple paths for the columns that are returned from the List and Get calls.

cbruno10 (changes requested):
@ParthaI Please see comments, thanks!

aws/table_aws_cloudtrail_query.go
ParthaI: We were using `v1.16.8` of Cloudtrail package, DescribeQuery API was not returning Delivery status, Delivery S3 URI attributes with this package. Now I have upgraded package to `v1.20.4` and added those missing columns.

cbruno10 (approved):

cbruno10:
@ParthaI @misraved I'm good with this PR, please resolve merge conflicts and merge when ready, thanks!

PR #1438, opened 2022-11-29
Fix usage of WithCache for getCommonColumns

kaidaguerre:
Fix the usage of `.WithCache` for `getCommonColumns`:
- declare single cached version of `getCommonColumns` -  `getCommonColumnsCached`
- declare function to return cache key which includes the region
- remove unnecessary caching from `getCallerIdentity`

### DIFF
--- a/aws/common_columns.go
+++ b/aws/common_columns.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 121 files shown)


cbruno10:
Closing this PR for now, as these changes and more are being committed to the `fix-cache` branch.

PR #1647, opened 2023-03-10
Fix column descriptions in `aws_vpc_security_group_rule` table. Closes #1644

### LINKED ISSUE #1644: Fix column descriptions in `aws_vpc_security_group_rule` table
**Describe the bug**
Descriptions for columns like `cidr_ipv4`, `cidr_ipv6` are incorrect, as they can only contain CIDRs, not referenced security groups. These should checked, along with all other columns in the table against descriptions provided in https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SecurityGroupRule.html.

v0.19.1

v0.95.0

**To reproduce**
View https://hub.steampipe.io/plugins/turbot/aws/tables/aws_vpc_security_group_rule or run `.inspect aws_vpc_security_group_rule`.

**Expected behavior**
Correct column descriptions

**Additional context**

karanpopat:


aws/table_aws_vpc_security_group_rule.go
```diff
@@ -106,18 +106,18 @@ func tableAwsVpcSecurityGroupRule(_ context.Context) *plugin.Table {
 			},
 			{
 				Name:        "cidr_ip",
-				Description: "[DEPRECATED] This column has been deprecated and will be removed in a future release, use cidr_ipv4 instead. The IPv4 CIDR range. It can be either a CIDR range or a source security group, not both. A single IPv4 address is denoted by /32 prefix length.",
```
cbruno10, line 109: This is an older column, based on the way we populate it, could it ever be a source security group, or was the description wrong back then too?

cbruno10 (changes requested):
@karanpopat Thanks for making these updates, I've added one question, please have a look when you get a chance.

cbruno10 (approved):

PR #1649, opened 2023-03-13
Add table aws_cloudwatch_metric_statistic_data_point

bigdatasourav:
# Integration test logs
<details>
  <summary>Logs</summary>

```
N/A
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
> select
  jsonb_pretty(dimensions) as dimensions,
  timestamp,
  round(minimum::numeric,2) as min_cpu,
  round(maximum::numeric,2) as max_cpu,
  round(average::numeric,2) as avg_cpu,
  sample_count
from
  aws_cloudwatch_metric_statistic_data_point
where
  namespace = 'AWS/EC2'
  and metric_name = 'CPUUtilization'
  and timestamp between '2023-03-15T00:00:00Z' and '2023-03-15T13:33:00Z'
  and period = 86400
  and dimensions = '[
    {"Name": "InstanceId", "Value": "i-0dd7043e0f6f0f36d"}
    ]'::jsonb
order by
  timestamp;
+----------------------------------------+---------------------------+---------+---------+---------+--------------+
| dimensions                             | timestamp                 | min_cpu | max_cpu | avg_cpu | sample_count |
+----------------------------------------+---------------------------+---------+---------+---------+--------------+
| [                                      | 2023-03-15T05:30:00+05:30 | 0       | 19.84   | 0.3     | 162          |
|     {                                  |                           |         |         |         |              |
|         "Name": "InstanceId",          |                           |         |         |         |              |
|         "Value": "i-0dd7043e0f6f0f36d" |                           |         |         |         |              |
|     }                                  |                           |         |         |         |              |
| ]                                      |                           |         |         |         |              |
+----------------------------------------+---------------------------+---------+---------+---------+--------------+

```
</details>


docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
+# Table: aws_cloudwatch_metric_statistic_data_point
+
+This table provides statistic data points for the specified metric.
+
+The maximum number of data points returned from a single call is 1,440. If you request more than 1,440 data points, CloudWatch returns an error. To reduce the number of data points, you can narrow the specified time range and make multiple requests across adjacent time ranges, or you can increase the specified period. Data points are not returned in chronological order.
+
+If you need to fetch more than 1440 data points then please use the `aws_cloudwatch_metric_data_point` table.
+
+- You **_must_** specify `metric_name`, `namespace`, `timestamp` and `dimensions` in a `where` clause in order to use this table. To fetch aggregate statistics data `dimensions` is not required.
```
cbruno10, line 9: We should exclude `dimensions` from the first sentence if we're going to contradict requiring it in the second sentence.

We should move the explanation about when to specify or not specify `dimensions` in another bullet

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+
+- You need to provide the timestamp value in a range like `timestamp >= '2023-03-15T00:00:00Z' and timestamp <= '2023-03-15T13:33:00Z' or like in the examples below.`
```
cbruno10, line 11: Can we update the table to default a start time to 24 hours ago and an end time to current time if we can't work it out based on quals? If so, we can then update the details here

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+
+- The GetMetricStatistics API used for this table cannot process multiple dimension values at a time, so you need to pass one set of dimensions to the query, like in the examples below.
```
cbruno10, line 13: What happens if you hit the API (or use the AWS CLI) and pass the same dimension name twice in a single request/command?

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+
+We recommend specifying the `period` column in the query to optimize the table output.
```
cbruno10, line 15: We should add details here on what happens if `period` is not specified.

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Aggregate CPU utilization of all ec2 instances for a given time frame
+
+```sql
+select
+  metric_name,
+  timestamp,
+  round(minimum::numeric,2) as min_cpu,
+  round(maximum::numeric,2) as max_cpu,
+  round(average::numeric,2) as avg_cpu,
```
cbruno10, line 27:
```suggestion
  round(minimum::numeric, 2) as min_cpu,
  round(maximum::numeric, 2) as max_cpu,
  round(average::numeric, 2) as avg_cpu,
```

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+  sample_count
+from
+  aws_cloudwatch_metric_statistic_data_point
+where
+  namespace = 'AWS/EC2'
+  and metric_name = 'CPUUtilization'
+  and timestamp between '2023-03-15T00:00:00Z' and '2023-03-15T13:33:00Z'
+  and period = 86400;
```
cbruno10, line 35: Should we add an `order by timestamp asc` to this query as well so the rows will be mapped oldest to newest?

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+  round(minimum::numeric,2) as min_cpu,
+  round(maximum::numeric,2) as max_cpu,
+  round(average::numeric,2) as avg_cpu,
+  sample_count
+from
+  aws_cloudwatch_metric_statistic_data_point
+where
+  namespace = 'AWS/EC2'
+  and metric_name = 'CPUUtilization'
+  and timestamp between '2023-03-15T00:00:00Z' and '2023-03-15T13:33:00Z'
```
cbruno10, line 34: Why is the timestamp over a 13 hour period if we give a period of 24 hours?

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+
+- You need to provide the timestamp value in a range like `timestamp >= '2023-03-15T00:00:00Z' and timestamp <= '2023-03-15T13:33:00Z' or like in the examples below.`
+
+- The GetMetricStatistics API used for this table cannot process multiple dimension values at a time, so you need to pass one set of dimensions to the query, like in the examples below.
+
+We recommend specifying the `period` column in the query to optimize the table output.
+
+## Examples
+
+### Aggregate CPU utilization of all ec2 instances for a given time frame
```
cbruno10, line 19:
```suggestion
### Aggregate CPU utilization of all EC2 instances
```

cbruno10: We can leave out the words `for a given time frame` since all data will be in a given time frame

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
cbruno10: Also, is a period of 24 hours OK here? What periods do people typically use when monitoring EC2 instance utilization? I feel like it'd be a smaller window (since instance CPU utilization can change quickly and frequently)

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+  sample_count
+from
+  aws_cloudwatch_metric_statistic_data_point
+where
+  namespace = 'AWS/EC2'
+  and metric_name = 'CPUUtilization'
+  and average > 80
+  and timestamp between '2023-03-15T00:00:00Z' and '2023-03-15T13:33:00Z'
+  and period = 86400
+  and dimensions = '[
```
cbruno10, line 56: Can you please use `jsonb_object` to build the JSONB object, like in https://hub.steampipe.io/plugins/turbot/net/tables/net_http_request#send-a-get-request-with-a-modified-user-agent, for this example and others?

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+from
+  aws_cloudwatch_metric_statistic_data_point
+where
+  namespace = 'AWS/EC2'
+  and metric_name = 'CPUUtilization'
+  and timestamp between '2023-03-15T00:00:00Z' and '2023-03-15T13:33:00Z'
+  and period = 86400;
+```
+
+### CPU average utilization of an ec2 instance over 80% for a given time frame
```
cbruno10, line 38: What does this query show? Is it meant to show data points (and thus the timestamps) when the average CPU utilization was over 80%?

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+
+```sql
+select
+  jsonb_pretty(dimensions) as dimensions,
+  timestamp,
+  minimum,
+  maximum,
+  average,
+  sum,
+  sample_count
```
cbruno10, line 73: Do we want to get all of the metrics for the purposes of this query? Which statistics are important/valuable?

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
cbruno10: Also, if we have a reasonable way to work out a default start/end time, could we make `timestamp` qual optional? So if a user only specifies `metric_name` and `namespace`, we could get results from the past 24 hours.

What period would we work out by default in that case?

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
```
cbruno10, line 176: Don't most of these statistics return the same data since the metric only has 1 data point per day per bucket? Which statistic should we use?

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
```diff
@@ -0,0 +1,190 @@
...
+
+- You **_must_** specify `metric_name`, `namespace`, `timestamp` and `dimensions` in a `where` clause in order to use this table. To fetch aggregate statistics data `dimensions` is not required.
+
+- You need to provide the timestamp value in a range like `timestamp >= '2023-03-15T00:00:00Z' and timestamp <= '2023-03-15T13:33:00Z' or like in the examples below.`
+
+- The GetMetricStatistics API used for this table cannot process multiple dimension values at a time, so you need to pass one set of dimensions to the query, like in the examples below.
+
+We recommend specifying the `period` column in the query to optimize the table output.
+
+## Examples
```
cbruno10, line 17: @bigdatasourav For all examples in here, please tune the timestamps, periods, and statistic columns we include the statement so the queries become more effective.

Also, we should add some examples showing a query that only passes in required quals

aws/table_aws_cloudwatch_metric_statistic_data_point.go
```diff
@@ -0,0 +1,173 @@
...
+					Name:    "namespace",
+					Require: plugin.Required,
+				},
+				{
+					Name:      "timestamp",
+					Operators: []string{">", ">=", "=", "<", "<="},
+					Require:   plugin.Required,
+				},
+				{
+					Name:    "period",
```
cbruno10, line 43: On any of the optional key quals, I think we need to add `CacheMatch: Exact`. For instance, if caching is enabled, if a user runs a query without any optional quals, and then runs a query with one or more optional quals specified,  they'll get unexpected results since the API should return 2 different data sets.

aws/table_aws_cloudwatch_metric_statistic_data_point.go
```diff
@@ -0,0 +1,173 @@
...
+			},
+			{
+				Name:        "dimensions",
+				Description: "The dimensions for the metric.",
+				Type:        proto.ColumnType_JSON,
+			},
+
+			// Steampipe standard columns
+			{
+				Name:        "title",
```
cbruno10, line 72: What are some example labels? Are they good titles?

aws/table_aws_cloudwatch_metric_statistic_data_point.go
```diff
@@ -0,0 +1,173 @@
...
+
+//// LIST FUNCTION
+
+func listCloudWatchMetricStatisticDataPoints(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+
+	// set the input parameters
+	params := &cloudwatch.GetMetricStatisticsInput{
+		MetricName: aws.String(d.EqualsQuals["metric_name"].GetStringValue()),
+		Namespace:  aws.String(d.EqualsQuals["namespace"].GetStringValue()),
+		Statistics: types.Statistic.Values(types.StatisticMaximum),
```
cbruno10, line 98: What value is in `types.StatisticMaximum`? Are we getting all statistics by passing in this input value?

aws/table_aws_cloudwatch_metric_statistic_data_point.go
```diff
@@ -0,0 +1,173 @@
...
+		case "<", "<=":
+			params.EndTime = aws.Time(time)
+		}
+	}
+
+	// set the period based on the duration between the start and end time
+	// https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudwatch@v1.25.1#GetMetricStatisticsInput.Period
+	duration := params.EndTime.Sub(*params.StartTime)
+	// if the duration is less than 3 hrs
+	if duration.Hours() < 3 {
```
cbruno10, line 119: @bigdatasourav Can you please add some comments showing examples of what the period would be and how many API results we can expect for a few different scenarios?

cbruno10 (changes requested):
@bigdatasourav Please see review suggestions/questions, thanks!

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: Updated the default end time to the current time and the default start time to 24hrs ago.
Made timestamp as an optional column, so if the user does not provide it, we can get the result from the last 24 hrs.
The default period will be 60 sec, fetching exactly 1440 datapoints.

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: updated the doc

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: The API returns an empty response if we pass another/same dimension in the query-
image
image

Same with the CLI as well.

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: updated in doc

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: updated the query

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: @cbruno10 if we use jsonb_object, we will get an error like - as it will not be able to unmarshal it to ]types.Dimension
```
Error: failed to unmarshal dimensions {
    "Name": "InstanceId",
    "Value": "i-0dd7043e0f6f0f36d"
}: json: cannot unmarshal object into Go value of type []types.Dimension (SQLSTATE HV000)
```

bigdatasourav: Is there any issue if we provide the value like this?
```
where
   dimensions = '[
    {"Name": "InstanceId", "Value": "i-0dd7043e0f6f0f36d"}
    ]';
```

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: yes

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: All are not required; updated the queries based on asked stat

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: Updated.

aws/table_aws_cloudwatch_metric_statistic_data_point.go
bigdatasourav: updated.

aws/table_aws_cloudwatch_metric_statistic_data_point.go
bigdatasourav: Labels are primarily based on resource id, and name or they can be the metric name as well. 

`i-0353536c53f7c8235 (test[sourav - test purpose] )`
`BucketSizeBytes`

aws/table_aws_cloudwatch_metric_statistic_data_point.go
bigdatasourav: `types.StatisticMaximum` holds the value `maximum`
If we pass any types.Statistic to the `types.Statistic.Values` function, in return you will get all.
```
// Values returns all known values for Statistic. Note that this can be expanded in
// the future, and so it is only as up to date as the client. The ordering of this
// slice is not guaranteed to be stable across updates.
func (Statistic) Values() []Statistic {
	return []Statistic{
		"SampleCount",
		"Average",
		"Sum",
		"Minimum",
		"Maximum",
	}
}
```

aws/table_aws_cloudwatch_metric_statistic_data_point.go
bigdatasourav: updated.

docs/tables/aws_cloudwatch_metric_statistic_data_point.md
bigdatasourav: updated.

aws/table_aws_cloudwatch_metric_statistic_data_point.go
cbruno10: Are those helpful titles when looking at rows? From the examples above, it's not clear to me that they are helpful titles out of the box.

aws/table_aws_cloudwatch_metric_statistic_data_point.go
```diff
@@ -0,0 +1,214 @@
...
+	// in this case 486000/1440 = 337, which is not multiple of 60, so the closest multiple of 60 after 337 is 360
+	// with this period we will get upto 1350 datapoints
+
+	duration := params.EndTime.Sub(*params.StartTime).Hours()
+	durationSec := int32(duration) * 3600
+	defaultPeriod := (int32(duration) * 3600) / 1440
+
+	// if the duration is less than 3 hrs
+	if duration < 3 {
+		params.Period = aws.Int32(10)
```
cbruno10, line 149: Why did we pick 10 seconds if under 3 hours? I couldn't find anything in AWS' docs, but I probably missed something

cbruno10 (changes requested):
@bigdatasourav please see additional comments, thanks!

aws/table_aws_cloudwatch_metric_statistic_data_point.go
```diff
@@ -0,0 +1,214 @@
...
+func tableAwsCloudWatchMetricStatisticDataPoint(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_cloudwatch_metric_statistic_data_point",
+		Description: "AWS CloudWatch Metric Statistic Data Point",
+		List: &plugin.ListConfig{
+			Hydrate: listCloudWatchMetricStatisticDataPoints,
+			IgnoreConfig: &plugin.IgnoreConfig{
+				ShouldIgnoreErrorFunc: shouldIgnoreErrors([]string{"InvalidParameterValue"}),
+			},
+			KeyColumns: []*plugin.KeyColumn{
```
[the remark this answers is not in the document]
bigdatasourav: Added it

aws/table_aws_cloudwatch_metric_statistic_data_point.go
```diff
@@ -0,0 +1,214 @@
...
+				params.StartTime = aws.Time(timestamp)
+				params.EndTime = aws.Time(timestamp)
+			case ">=", ">":
+				params.StartTime = aws.Time(timestamp)
+			case "<", "<=":
+				params.EndTime = aws.Time(timestamp)
+			}
+		}
+	} else {
+		params.StartTime = aws.Time(time.Now().AddDate(0, 0, -1))
```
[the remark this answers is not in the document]
bigdatasourav: updated

aws/table_aws_cloudwatch_metric_statistic_data_point.go
bigdatasourav: updated the logic, default minimum is 60 sec

roisec:
there is any expectation when it will ready?

misraved (approved):

misraved (approved):

PR #1670, opened 2023-03-23
Add table aws_ssm_document_permission. closes #1669

### LINKED ISSUE #1669: Add table aws_ssm_document_permission
**References**
https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_DescribeDocumentPermission.html

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
N/A
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
> select * from aws_ssm_document_permission where document_name = 'test';
+---------------+-------------------+-------------------------+---------------------------------+----------+-----------+-----------+--------------+---------------------------+
| document_name | shared_account_id | shared_document_version | account_ids                     | title    | partition | region    | account_id   | _ctx                      |
+---------------+-------------------+-------------------------+---------------------------------+----------+-----------+-----------+--------------+---------------------------+
| test          | 321321312412      | $DEFAULT                | ["123456789012","321321312412"] | $DEFAULT | aws       | us-east-1 | 367493749284 | {"connection_name":"aws"} |
| test          | 123456789012      | $DEFAULT                | ["123456789012","321321312412"] | $DEFAULT | aws       | us-east-1 | 367493749284 | {"connection_name":"aws"} |
+---------------+-------------------+-------------------------+---------------------------------+----------+-----------+-----------+--------------+---------------------------+
```
</details>

### DIFF
--- a/aws/table_aws_ssm_document_permission.go
+++ b/aws/table_aws_ssm_document_permission.go
@@ -1,0 +1,133 @@
+package aws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/service/ssm"
+	"github.com/aws/aws-sdk-go-v2/service/ssm/types"
+
+	ssmv1 "github.com/aws/aws-sdk-go/service/ssm"
+
+	"github.com/turbot/steampipe-plugin-sdk/v5/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin/transform"
+)
+
+func tableAwsSSMDocumentPermission(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_ssm_document_permission",
+		Description: "AWS SSM Document Permission",
+		List: &plugin.ListConfig{
+			Hydrate: listAwsSSMDocumentPermissions,
+			KeyColumns: []*plugin.KeyColumn{
+				{Name: "document_name", Require: plugin.Required},
+			},
+		},
+		GetMatrixItemFunc: SupportedRegionMatrix(ssmv1.EndpointsID),
+		Columns: awsRegionalColumns([]*plugin.Column{
+			{
+				Name:        "document_name",
+				Description: "The name of the Systems Manager document.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "shared_account_id",
+				Description: "The Amazon Web Services account ID where the current document is shared.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("AccountSharingInfo.AccountId"),
+			},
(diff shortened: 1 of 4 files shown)


aws/table_aws_ssm_document_permission.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin/transform"
+)
+
+func tableAwsSSMDocumentPermission(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_ssm_document_permission",
+		Description: "AWS SSM Document Permission",
+		List: &plugin.ListConfig{
+			Hydrate: listAwsSSMDocumentPermissions,
```
cbruno10, line 21: Should this have a required key qual, or could we use parent hydrate? How many docs are common in a region?

aws/table_aws_ssm_document_permission.go
```diff
@@ -0,0 +1,133 @@
...
+			},
+			{
+				Name:        "shared_document_version",
+				Description: "The version of the current document shared with the account.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("AccountSharingInfo.SharedDocumentVersion"),
+			},
+			{
+				Name:        "account_ids",
+				Description: "The account IDs that have permission to use this document.The ID can be either an AWS account or All.",
```
misraved, line 47:
```suggestion
				Description: "The account IDs that have permission to use this document. The ID can be either an AWS account or All.",
```

aws/table_aws_ssm_document_permission.go
```diff
@@ -0,0 +1,133 @@
...
+				Type:        proto.ColumnType_JSON,
+			},
+
+			// Steampipe standard columns
```
misraved, line 51: @ParthaI Is there a specific reason why we did not add other standard columns?

misraved (commented):
@ParthaI could you please take a look at the review comments? Thanks!!

aws/table_aws_ssm_document_permission.go
ParthaI: @cbruno10 My first thought was to use parent hydrate for this table, but it took over 25 minutes to get the result. There are around 15000 parent resources. Therefore I substituted needed quals for parent hydrate.

aws/table_aws_ssm_document_permission.go
ParthaI: @misraved At first AWS does not treat `aws_ssm_document_permission` as a resource type.
As per the steampipe plugin table development standard if any hydrate data supports pagination we considered it as a separate table. Also the API response doesn't contain any tag information. So that I have not added any other standard column

PR #1703, opened 2023-04-12
Adding source_account_id column to aws_securityhub_finding table

gabrielsoltz:
The table `aws_securityhub_finding` exposes a column `account_id` which based on this document: https://hub.steampipe.io/plugins/turbot/aws/tables/aws_securityhub_finding is "The AWS Account ID in which the resource is located.". This can be very confusing, in this case, account_id is where security hub is running, not where the affected resource by a security hub finding is running. When you are aggregating findings from different accounts, is important to understand or been able to filter by the affected account: where the affected resource lives.

That's why i'm adding the column `affected_account_id`

### DIFF
--- a/aws/table_aws_securityhub_finding.go
+++ b/aws/table_aws_securityhub_finding.go
@@ -43,7 +43,7 @@ func tableAwsSecurityHubFinding(_ context.Context) *plugin.Table {
 				{Name: "verification_state", Require: plugin.Optional, Operators: []string{"=", "<>"}},
 				{Name: "workflow_state", Require: plugin.Optional, Operators: []string{"=", "<>"}},
 				{Name: "workflow_status", Require: plugin.Optional, Operators: []string{"=", "<>"}},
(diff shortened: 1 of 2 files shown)


misraved:
Thanks @gabrielsoltz for adding the new column 👍.

Could you please add an example query in table doc file which reflects the usage of this new column?

misraved:
@gabrielsoltz the changes look good and thanks for adding the example queries 👍.

Do you think `source_account_id` would be a better fit for this column? From the description mentioned in https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_AwsSecurityFindingFilters.html#securityhub-Type-AwsSecurityFindingFilters-AwsAccountId, I feel that source_account_id fits the bill since it reflects which account feeds into the findings.

gabrielsoltz:
@misraved works for me, changed :)

misraved (approved):

misraved (approved):

PR #1718, opened 2023-05-02
Add table aws_config_retention_configuration closes #1705

### LINKED ISSUE #1705: Add table aws_config_retention_configuration
I believe there is no way so far to know the retention of the AWS Config: https://docs.aws.amazon.com/config/latest/developerguide/delete-config-data-with-retention-period.html

I believe the most elegant way would be to have a column named something like "retention_period", integer type, with the number of days of retention. API spec: https://docs.aws.amazon.com/config/latest/APIReference/API_DescribeRetentionConfigurations.html#API_DescribeRetentionConfigurations_RequestSyntax 

none

**Additional context**
In checking the compliance check https://hub.steampipe.io/mods/turbot/aws_compliance/controls/control.config_enabled_all_regions I noticed that if the collection is enabled but retention is set to 30 days, check is still green which I find incomplete.

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
N/A
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
> select
   name,
  retention_period_in_days,
  title
from
  aws_config_retention_configuration
where
  retention_period_in_days < 356;
+---------+--------------------------+---------+
| name    | retention_period_in_days | title   |
+---------+--------------------------+---------+
| default | 225                      | default |
+---------+--------------------------+---------+

> select * from aws_config_retention_configuration where name = 'default'
+---------+--------------------------+---------+-----------+-----------+--------------+---------------------------+
| name    | retention_period_in_days | title   | partition | region    | account_id   | _ctx                      |
+---------+--------------------------+---------+-----------+-----------+--------------+---------------------------+
| default | 225                      | default | aws       | us-east-1 | 123456789012 | {"connection_name":"aws"} |
+---------+--------------------------+---------+-----------+-----------+--------------+---------------------------+
```
</details>


misraved (commented):
@ParthaI I think we should remove the `GetConfig` from the table since the `default` retention configuration is unique to a region and it will give an error when queried.

misraved (commented):
@ParthaI I think we should remove the `GetConfig` from the table since the `default` retention configuration is only unique to a region and it will give an error when queried.

misraved (approved):

PR #1737, opened 2023-05-16
Fix source_account_id optional qual column in aws_security_hub_finding table 

gabrielsoltz:
Field source_account_id was incorrectly defined, fixed.


misraved:
Good catch @gabrielsoltz, apologies for the earlier miss 👍.

misraved (approved):

PR #1869, opened 2023-08-07
nlbs do not use sec groups

tinder-tder:
Documentation update, AWS NLBs do not use security group attachments

### DIFF
--- a/docs/tables/aws_ec2_network_load_balancer.md
+++ b/docs/tables/aws_ec2_network_load_balancer.md
@@ -4,17 +4,6 @@ AWS Network Load Balancer (NLB) distributes end user traffic across multiple clo
 
 ## Examples
 
-### Security group attached to the network load balancer
-
-```sql
-select
-  name,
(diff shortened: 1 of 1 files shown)


rajlearner17:
@tinder-tder Thanks for the PR. Good catch! 👍 

The query will always return an empty row. So `security_groups` column may not add any value to the table. Not sure why we added this 🤔

rajlearner17 (approved):
Makes sense to take off this example query.

misraved:
Nice catch @tinder-tder, could you please raise an issue or a PR (we love PRs 👍 ) to remove the associated `security_groups` column?

tinder-tder:
> @tinder-tder Thanks for the PR. Good catch! 👍
> 
> The query will always return an empty row. So `security_groups` column may not add any value to the table. Not sure why we added this 🤔 . Worth taking it off, then.

I'm guessing it was a copy/mod of the alb table code/docs

PR #1873, opened 2023-08-11
Fixed better error message when AWS creds resolver fails Closes #1846

### LINKED ISSUE #1846: Add better error message when AWS creds resolver fails
When I pass in missing credentials, e.g., specify a profile in `aws.spc` that doesn't exist in my AWS config or credentials file, the plugin returns an error:
```
Error: rpc error: code = DeadlineExceeded desc = operation error S3: ListBuckets, failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, request canceled, context deadline exceeded (SQLSTATE HV000)
```

This is most likely due to the way AWS SDK Go v2 resolves creds, https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specifying-credentials, and doesn't error when a missing profile is provided to it when setting up the shared config.

The plugin should provide a better error message to point me in the right direction that my creds are missing or not specified

n/A

**Additional context**

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
N/A
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
> select * from aws_account

Error: connection config has an incorrect aws profile 'aws_nagrajaaa', update the configuration file (SQLSTATE HV000)

+-----------------+-----+-----------------+------------------+--------------------------+---------------------------------+-----------------------------------+--------------------------------+-------------------------------------+-------+------+-----------+--------+--->
| account_aliases | arn | organization_id | organization_arn | organization_feature_set | organization_master_account_arn | organization_master_account_email | organization_master_account_id | organization_available_policy_types | title | akas | partition | region | ac>
+-----------------+-----+-----------------+------------------+--------------------------+---------------------------------+-----------------------------------+--------------------------------+-------------------------------------+-------+------+-----------+--------+--->
+-----------------+-----+-----------------+------------------+--------------------------+---------------------------------+-----------------------------------+--------------------------------+-------------------------------------+-------+------+-----------+--------+--->

```
</details>


aws/service.go
```diff
@@ -1704,7 +1704,16 @@ func getBaseClientForAccountUncached(ctx context.Context, d *plugin.QueryData, _
 	if awsSpcConfig.Profile != nil {
 		profile := aws.ToString(awsSpcConfig.Profile)
 		plugin.Logger(ctx).Trace("getBaseClientForAccountUncached", "connection_name", d.Connection.Name, "status", "profile_found", "profile", profile)
+
+		// Check connection config has a valid aws profile
+		cfg, _ := config.LoadDefaultConfig(ctx, config.WithSharedConfigProfile(profile))
+		_, err := cfg.Credentials.Retrieve(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("connection config has an incorrect aws profile '%v', update the configuration file", profile)
```
misraved, line 1712:
```suggestion
			return nil, fmt.Errorf("connection config has an invalid AWS profile '%v', Please update the configuration file", profile)
```

misraved:
Verified that the changes also work for aggregator profiles - 
```
 select * from aws_all.aws_account

Error: connection config has an invalid AWS profile 'default23', Please update the configuration file
	connection config has an invalid AWS profile 'testaab12', Please update the configuration file
```

cbruno10:
@ParthaI :
- Does this work if I pass a non-existent profile name in `AWS_PROFILE` env var?
- Are we able to tell the user which connection is causing the issue?
- What does `invalid profile` mean (per the error message)? Is it always if a non-existent profile name is given? Or could it be if a profile definition in the AWS credentials/config file has a missing or incorrect property, incorrect data, etc.?
- We should remove the text `Please update the configuration file.` from the error message, it's unclear as to what update is required and in general we don't really use that in our error messages.
- Is this the best place in the code flow to check?
- Are the default config loading and credential retrieval function calls expensive?
- When we check the results of `_, err := cfg.Credentials.Retrieve(ctx)`, would that credentials retrieval be helpful for any other auth methods?

ParthaI:
@cbruno10, Here are the few key points as per the above comments, please take a look, Thanks!

Q: Will this function work if I provide a non-existent profile name in the AWS_PROFILE environment variable?
A: No, it will not work if a non-existent profile name is provided in the AWS_PROFILE environment variable.

Q: Can we identify the specific connection that is causing the problem for the user?
A: Yes, we can append the connection name to the error message.

Q: What does "invalid profile" mean according to the error message? Is it only when a non-existent profile name is used? Or could it also refer to missing or incorrect properties, improper data, etc., within a profile definition in the AWS credentials/config file?
We should exclude the phrase "Please update the configuration file." from the error message. It's unclear what specific update is required, and this phrase isn't commonly used in our error messages.
A: "Invalid profile" refers specifically to a non-existent profile name being used. The code block `cfg, err := config.LoadDefaultConfig(ctx, config.WithSharedConfigProfile(profile))` always generates an error when a non-existent profile name is provided. To catch this, we used `_, err := cfg.Credentials.Retrieve(ctx)`.

Q: Is this the most appropriate point in the code to perform this check?
A: I believe it's the most suitable location to handle this check. Other authentication errors aren't detected by the `cfg.Credentials.Retrieve()` function.

Q: Do the calls to load default configuration and retrieve credentials impose a significant performance overhead?
A: No, these calls do not result in any additional costs.

Q: When we evaluate the results of `_, err := cfg.Credentials.Retrieve(ctx)`, could this credentials retrieval be beneficial for other authentication methods?
A: The `Retrieve()` function provides a structure (https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v1.19.0/aws#Credentials) that includes `Access Key`, `Secret key`, `Session Token`, `Source`, `CanExpire`, and `Expires`. However, this information may not be useful for authentication methods such as AssumeRole Credentials (With MFA), AssumeRole Credentials (in ECS), AWS-Vault Credentials, and Credentials from an EC2 Instance Profile.

PR #1892, opened 2023-09-07
Add table aws_batch_job. Closes#1891

madhushreeray30:
# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
```
</details>

### DIFF
--- a/aws/table_aws_batch_job.go
+++ b/aws/table_aws_batch_job.go
@@ -1,0 +1,312 @@
+package aws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/batch"
+	// "github.com/aws/aws-sdk-go-v2/service/batch/types"
+
+	batchv1 "github.com/aws/aws-sdk-go/service/batch"
+
+	"github.com/turbot/steampipe-plugin-sdk/v5/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin/transform"
+)
+
+//// TABLE DEFINITION
+
+func tableAwsBatchJob(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_batch_job",
+		Description: "AWS Batch Job",
+		Get: &plugin.GetConfig{
+			KeyColumns: plugin.SingleColumn("name"),
+			IgnoreConfig: &plugin.IgnoreConfig{
+				ShouldIgnoreErrorFunc: shouldIgnoreErrors([]string{"InvalidParameterValueException"}),
+			},
+			Hydrate: getBatchJob,
+		},
+		List: &plugin.ListConfig{
+			Hydrate: listAwsBatchJobs,
+		},
+		GetMatrixItemFunc: SupportedRegionMatrix(batchv1.EndpointsID),
+		Columns: awsRegionalColumns([]*plugin.Column{
+			{
+				Name:        "arn",
+				Description: "The Amazon Resource Name (ARN) of the job.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("JobArn"),
+			},
+			{
+				Name:        "job_id",
+				Description: "The job ID.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "name",
+				Description: "The job name.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("JobName"),
+			},
+			{
+				Name:        "array_properties",
+				Description: "The array properties of the job, if it's an array job.",
+				Type:        proto.ColumnType_JSON,
+			},
+			{
+				Name:        "container",
+				Description: "An object that represents the details of the container that's associated with the job.",
+				Type:        proto.ColumnType_JSON,
+			},
+			{
+				Name:        "created_at",
+				Description: "The Unix timestamp (in milliseconds) for when the job was created.",
+				Type:        proto.ColumnType_TIMESTAMP,
+			},
+			{
+				Name:        "job_definition",
+				Description: "The Amazon Resource Name (ARN) of the job definition.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "single_node_properties",
+				Description: "The node properties for a single node in a job summary list.",
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("NodeProperties"),
+			},
+			{
+				Name:        "started_at",
+				Description: "The Unix timestamp for when the job was started. More specifically, it's when the job transitioned from the STARTING state to the RUNNING state.",
+				Type:        proto.ColumnType_TIMESTAMP,
+			},
+			{
+				Name:        "status",
+				Description: "The current status for the job.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "status_reason",
+				Description: "A short, human-readable string to provide more details for the current status of the job.",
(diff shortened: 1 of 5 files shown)


aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,312 @@
+package aws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/batch"
+	// "github.com/aws/aws-sdk-go-v2/service/batch/types"
```
ParthaI, line 8:
```suggestion
```

aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,312 @@
...
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("JobArn"),
+			},
+			{
+				Name:        "job_id",
+				Description: "The job ID.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "name",
```
ParthaI, line 47: Can you please rename the column name to `job_name` instead? We should keep the column names as per the API response except the ARN column.

aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,312 @@
...
+				Name:        "attempts",
+				Description: "A list of job attempts that are associated with this job.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     getBatchJob,
+			},
+			{
+				Name:        "depends_on",
+				Description: "A list of job IDs that this job depends on.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     getBatchJob,
```
ParthaI, line 114: It would be better to place all the JSON columns all together.

aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,312 @@
...
+	// Limiting the results
+	maxLimit := int32(1000)
+	if d.QueryContext.Limit != nil {
+		limit := int32(*d.QueryContext.Limit)
+		if limit < maxLimit {
+			if limit < 1 {
+				maxLimit = 1
+			} else {
+				maxLimit = limit
+			}
```
ParthaI, line 242:
```suggestion
			maxLimit = limit
```

aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,312 @@
...
+		plugin.Logger(ctx).Error("aws_batch_job.listAwsBatchJobs", "connection_error", err)
+		return nil, err
+	}
+	if svc == nil {
+		// Unsupported region, return no data
+		return nil, nil
+	}
+
+	// Limiting the results
+	maxLimit := int32(1000)
```
ParthaI, line 234: Please update the maxLimit value to `100` as per the [API documentation](https://docs.aws.amazon.com/batch/latest/APIReference/API_ListJobs.html#API_ListJobs_RequestSyntax).

aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,312 @@
...
+		Description: "AWS Batch Job",
+		Get: &plugin.GetConfig{
+			KeyColumns: plugin.SingleColumn("name"),
+			IgnoreConfig: &plugin.IgnoreConfig{
+				ShouldIgnoreErrorFunc: shouldIgnoreErrors([]string{"InvalidParameterValueException"}),
+			},
+			Hydrate: getBatchJob,
+		},
+		List: &plugin.ListConfig{
+			Hydrate: listAwsBatchJobs,
```
ParthaI, line 31: Could you please add the optional key quals for minimizing the result and the API time consumption?

We can filter the results by providing [these values](https://docs.aws.amazon.com/batch/latest/APIReference/API_ListJobs.html#API_ListJobs_RequestSyntax) as an input parameter of API call

aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+	}
+
+	input := &batch.ListJobsInput{
+		MaxResults: aws.Int32(maxLimit),
+	}
+
+	pagesLeft := true
+
+	for pagesLeft {
```
ParthaI, line 252: Instead of performing the manual pagination, can you please update the logic to use [SDK pagination operation](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/batch@v1.26.5#NewListJobsPaginator)?

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.21
 
 require (
 	github.com/aws/aws-sdk-go v1.44.189
-	github.com/aws/aws-sdk-go-v2 v1.19.0
+	github.com/aws/aws-sdk-go-v2 v1.21.0
```
ParthaI, line 7: @madhushreeray30, we are updating the AWS SDK version `1.19.0` to `1.21.0`. Are there any significant changes or breaking changes between these two versions that we should be aware of?

ParthaI (commented):
@madhushreeray30 I have left a few review comments, please take a look thanks!

Also please cross-check the PR why the lint is failing for it.

aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,331 @@
...
+			IgnoreConfig: &plugin.IgnoreConfig{
+				ShouldIgnoreErrorFunc: shouldIgnoreErrors([]string{"InvalidParameterValueException"}),
+			},
+			Hydrate: getBatchJob,
+			Tags:    map[string]string{"service": "batch", "action": "DescribeJobs"},
+		},
+		List: &plugin.ListConfig{
+			Hydrate: listAwsBatchJobs,
+			Tags:    map[string]string{"service": "batch", "action": "ListJobs"},
+			KeyColumns: []*plugin.KeyColumn{
```
ParthaI, line 33: Could you please add the columns job_name, job_definition, and created_at as optional quals? The filter input param supports these values for filtering out the result.

created_at optional column should support the operator(=, <, >, <=, >=), based on the operator we can use the filter input `BEFORE_CREATED_AT` and `AFTER_CREATED_AT`

Note: Please update the table doc for the statement `Only one filter can be used at a time. When the filter is used, jobStatus is ignored`. I am not sure though how the API behaves.

aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+
+	// Get call
+	data, err := svc.DescribeJobs(ctx, params)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_batch_job.getBatchJob", "api_error", err)
+		return nil, err
+	}
+
+	return data.Jobs, nil
```
ParthaI, line 330: We should return to a single Job instead of an array of Jobs. The [DescripeJobs](https://docs.aws.amazon.com/batch/latest/APIReference/API_DescribeJobs.html) API returns a list of jobs as a response.

aws/table_aws_batch_job.go
```diff
@@ -0,0 +1,331 @@
...
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "job_name",
+				Description: "The job name.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("JobName"),
+			},
+			{
+				Name:        "created_at",
```
ParthaI, line 74: We may need a transform function(for converting the UNIX timestamp to our expected format) for this column value to be populated. We are getting the column value as number "[createdAt](https://docs.aws.amazon.com/batch/latest/APIReference/API_JobSummary.html#Batch-Type-JobSummary-createdAt)". 

Please check the same for other columns if we get the Unix timestamp value for a timestamp column type.

ParthaI (commented):
@madhushreeray30. Please take a look at the review comments. Thanks!

Could you please add the doc for this table?

PR #1927, opened 2023-10-01
fix table_aws_securityhub_finding field source_account_id

gabrielsoltz:
The field `source_account_id` in table `table_aws_securityhub_finding` was trying to fetch that field, which doesn't exist in the AWS API. The correct API field for this table field is `AwsAccountId`, so I'm adding the `Transform` option. 

Cheers 🥂 

Before: 
```
+---------------------------------------------------+-------------------+
| title                                             | source_account_id |
+---------------------------------------------------+-------------------+
| S3 bucket server access logging should be enabled | <null> |
```

After:
```
+---------------------------------------------------+-------------------+
| title                                             | source_account_id |
+---------------------------------------------------+-------------------+
| S3 bucket server access logging should be enabled | 012345678901      |
```

### DIFF
--- a/aws/table_aws_securityhub_finding.go
+++ b/aws/table_aws_securityhub_finding.go
@@ -249,6 +249,7 @@ func tableAwsSecurityHubFinding(_ context.Context) *plugin.Table {
 				Name:        "source_account_id",
 				Description: "The account id where the affected resource lives.",
 				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("AwsAccountId"),
 			},
 			/// Steampipe standard columns
 			{


cbruno10:
Hey @gabrielsoltz , I found that this column was originally added in https://github.com/turbot/steampipe-plugin-aws/pull/1703 (by you actually 😄 ). When you added this column, do you recall if the column had the correct values back then

gabrielsoltz:
Hi @cbruno10, yes, guilty. It was not working. I was using a local plugin version and forgot to fix it here. Looks like I'm the only one using this table. :p

misraved (approved):

PR #2058, opened 2024-01-19
Removed the table aws_organizations_organizational_unit Closes # 2057

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
```
</details>


aws/table_aws_organizations_account.go
```diff
@@ -122,9 +122,8 @@ func listOrganizationsAccounts(ctx context.Context, d *plugin.QueryData, _ *plug
 		return nil, err
 	}
 
-	// The maximum number for MaxResults parameter is not defined by the API
-	// We have set the MaxResults to 1000 based on our test
-	maxItems := int32(1000)
+	// Limit the result
+	maxItems := int32(20)
```
cbruno10, line 126: Why are we lowering this?

aws/plugin.go
```diff
@@ -334,7 +334,6 @@ func Plugin(ctx context.Context) *plugin.Plugin {
 			"aws_oam_sink":                                                 tableAwsOAMSink(ctx),
 			"aws_opensearch_domain":                                        tableAwsOpenSearchDomain(ctx),
 			"aws_organizations_account":                                    tableAwsOrganizationsAccount(ctx),
-			"aws_organizations_organizational_unit":                        tableAwsOrganizationsOrganizationalUnit(ctx),
```
cbruno10, line 337: Can we comment this out instead?

cbruno10 (changes requested):
@ParthaI Please see review comments, thanks!

aws/table_aws_organizations_account.go
ParthaI: As per AWS [API doc](https://docs.aws.amazon.com/organizations/latest/APIReference/API_ListAccounts.html) it is `20`. Earlier it was not mentioned in the doc so we had added it as per our testing. If we pass the max result value to `1000` we encounter the following error.

```
Error: aws: operation error Organizations: ListAccounts, https response error StatusCode: 400, RequestID: 5f591f02-1a55-419c-ab42-7f46dbcfc5cc, InvalidInputException: You provided a value that exceeds the maximum allowed.
```

aws/plugin.go
ParthaI: I added back the removed code and commented it.

cbruno10 (approved):

PR #2072, opened 2024-02-12
Fixed Interface Conversion Error When Querying aws.aws_health_affected_entity Closes #2069

### LINKED ISSUE #2069: Interface Conversion Error When Querying aws.aws_health_affected_entity
**Describe the bug**
When I run a query on the `aws.aws_health_affected_entity` table, I get this error below.  It happens when there are events present (there are records returned from `aws.aws_health_event`), otherwise it returns 0 records.

```
ERROR: rpc error: code = Unknown desc = aws: interface conversion: interface {} is types.Event, not *types.Event
```

v0.21.4

0.130.0

**To reproduce**
Run `select * from aws.aws_health_affected_entity` when `select * from aws.aws_health_event` returns results/events.

**Expected behavior**
It should return results instead of throwing the error mentioned above.

**Additional context**
I am querying AWS accounts that are in a gov-cloud region.

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
N/A
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
```
</details>

### DIFF
--- a/aws/table_aws_health_affected_entity.go
+++ b/aws/table_aws_health_affected_entity.go
@@ -87,7 +87,7 @@ func tableAwsHealthAffectedEntity(_ context.Context) *plugin.Table {
 
 func listHealthAffectedEntities(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
 	// Create Session
-	event := h.Item.(*types.Event)
+	event := h.Item.(types.Event)
 
 	// Limiting the results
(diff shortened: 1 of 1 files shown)


ParthaI:
We need to do proper testing for the code changes. Due to the lack of access in our environment, the table throws `SubscriptionRequiredException` error, so it is in draft mode.

misraved (approved):

PR #2189, opened 2024-05-14
Fixed failed to populate column 'akas' in aws_health_affected_entity, panic runtime error 😱 Closes #2183

### LINKED ISSUE #2183: Failed to populate column 'akas' in aws_health_affected_entity, panic runtime error 😱
**Describe the bug**
When I `select * from aws.aws_health_affected_entity` I get the error:
```
ERROR:  rpc error: code = Unknown desc = aws: failed to populate column 'akas': rpc error: code = Internal desc = transform EnsureStringArray failed with panic runtime error: invalid memory address or nil pointer dereference 
```

Example: v0.22.2

Example: v0.137.0

**To reproduce**
When I `select * from aws.aws_health_affected_entity` I get the error:
```
ERROR:  rpc error: code = Unknown desc = aws: failed to populate column 'akas': rpc error: code = Internal desc = transform EnsureStringArray failed with panic runtime error: invalid memory address or nil pointer dereference 
```

**Expected behavior**
It should return some results, because I have 21 records from `select * from aws.aws_health_event` with lots of open and upcoming events.

**Additional context**
I am querying against an AWS gov-cloud account.

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
N/A
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```

```
</details>


aws/table_aws_health_event.go
```diff
@@ -91,7 +92,7 @@ func tableAwsHealthEvent(_ context.Context) *plugin.Table {
 				Name:        "akas",
 				Description: resourceInterfaceDescription("akas"),
 				Type:        proto.ColumnType_JSON,
-				Transform:   transform.FromField("Arn").Transform(transform.EnsureStringArray),
+				Transform:   transform.FromField("Arn").Transform(getHealthEventAka),
```
misraved, line 95: @ParthaI I do not think we need to add the new Transform function here.

aws/table_aws_health_event.go
```diff
@@ -160,6 +161,29 @@ func listHealthEvents(ctx context.Context, d *plugin.QueryData, h *plugin.Hydrat
...
+			}
+			return []string{*v}, nil
+		default:
+			str := fmt.Sprintf("%v", d.Value)
+			return []string{str}, nil
+		}
+
+	}
+	return nil, nil
+}
```
misraved, line 185: I don't think this is required either because `event_arn` will never be nil.

misraved (approved):

PR #2198, opened 2024-05-22
Add aws_codestar_notificaton_rule table

gertvv:
I recently discovered this project and it is a life saver! I was missing the ability to query CodeStar notification rules (used e.g. to set up notifications of CodePipeline build status) so had a go at implementing this.

This is my first time working with Go so I don't expect this PR is ready to be merged and at this point could use some input - e.g. what tests to add, perhaps how to reshape some of the columns (e.g. targets), and what to do about the tags.

# Integration test logs

I haven't run these yet but would be happy to if you think this PR is worth developing further

# Example query results
<details>
  <summary>Results</summary>

```
select * from aws_codestar_notification_rule;

-[ RECORD 1  ]---------------------------------------------------------------------------
arn                     | arn:aws:codestar-notifications:eu-central-1:***:notificationrule/1725915dbd009519cd82707d18d8506e5f5d2926
name                    | some-pipeline-notification
resource                | arn:aws:codepipeline:eu-central-1:***:some-pipeline
detail_type             | FULL
status                  | ENABLED
event_types             | [{"EventTypeId":"codepipeline-pipeline-pipeline-execution-succeeded","EventTypeName":"Pipeline execution: Succeeded","ResourceType":"Pipeline","ServiceName":"CodePipeline"},{"EventTypeId":"codepipeline-pipeline-stage-execution-canceled","EventTypeName":"Stage execution: Canceled","ResourceType":"Pipeline","ServiceName":"CodePipeline"},{"EventTypeId":"codepipeline-pipeline-action-execution-failed","EventTypeName":"Action execution: Failed","ResourceType":"Pipeline","ServiceName":"CodePipeline"},{"EventTypeId":"codepipeline-pipeline-manual-approval-needed","EventTypeName":"Manual approval: Needed","ResourceType":"Pipeline","ServiceName":"CodePipeline"}]
targets                 | [{"TargetAddress":"arn:aws:sns:eu-central-1:***:my-topic","TargetStatus":"ACTIVE","TargetType":"SNS"}]
created_by              | ***
created_timestamp       | 2024-05-21T22:13:57+02:00
last_modified_timestamp | 2024-05-21T22:13:57+02:00
tags_src                | {}
partition               | aws
region                  | eu-central-1
account_id              | ***
sp_connection_name      | aws
sp_ctx                  | {"connection_name":"aws","steampipe":{"sdk_version":"5.10.0"}}
_ctx                    | {"connection_name":"aws","steampipe":{"sdk_version":"5.10.0"}}


select rule.name as rule_name, rule.resource, target->>'TargetAddress' as target, topic.title as topic
from aws_codestar_notification_rule as rule
cross join jsonb_array_elements(rule.targets) target
left join aws_sns_topic as topic on target->>'TargetAddress' = topic.topic_arn;

-[ RECORD 1  ]---------------------------------------------------------------------------
rule_name | some-pipeline-notification
resource  | arn:aws:codepipeline:eu-central-1:***:some-pipeline
target    | arn:aws:sns:eu-central-1:***:my-topic
topic     | my-topic

```
</details>


aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,188 @@
...
+				Description: "The Amazon Resource Name (ARN) of the notification rule.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("Arn"),
+			},
+			{
+				Name:        "name",
+				Description: "The name of the notification rule.",
+				Type:        proto.ColumnType_STRING,
+				Hydrate:     describeNotificationRule,
+				Transform:   transform.FromField("Name"),
```
misraved, line 48:
```suggestion
```
Transform function isn't needed since `FromField` transformation is applied by default 👍 !!

@gertvv you can remove all the references of `Transform:   transform.FromField` where by default the API returns data in upper camel case.

aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,188 @@
...
+			{
+				Name:        "tags_src",
+				Description: "The tags associated with the notification rule.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     describeNotificationRule,
+				Transform:   transform.FromField("Tags"),
+			},
+		}),
+	}
+}
```
misraved, line 115: @gertvv I think the column definition is missing the standard columns like `akas`, `title` and `tags` columns.

Please refer https://github.com/turbot/steampipe-plugin-aws/blob/main/aws/table_aws_ec2_instance.go#L488-L506 for additional information.

misraved (commented):
@gertvv thank you so much for the contribution 🎉. The table structure along with the API usage looks great.

I have added a few review comments that you can start working on. Please do add the table docs with example queries that best reflect the table usage.

gertvv:
Excellent, thanks for the comments! I have some travel coming up but will try to update the PR when I can.

gertvv:
I rebased the original commit to main and implemented the suggestions. I hope the choice of "Name" for title and "Arn" for akas makes sense.

aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,193 @@
...
+//// TABLE DEFINITION
+
+func tableAwsCodestarNotificationRule(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_codestar_notification_rule",
+		Description: "AWS CodeStar Notification Rule",
+		Get: &plugin.GetConfig{
+			KeyColumns: plugin.SingleColumn("arn"),
+			IgnoreConfig: &plugin.IgnoreConfig{
+				ShouldIgnoreErrorFunc: shouldIgnoreErrors([]string{"NotFound", "InvalidParameter"}),
```
ParthaI, line 25: Hi @gertvv, should we consider handling the `InvalidParameter` exception here? I believe it would be beneficial to inform the user that the provided value is invalid, allowing them to correct it. What are your thoughts on this?


Additionally, the API [DescribeNotificationRule](https://docs.aws.amazon.com/codestar-notifications/latest/APIReference/API_DescribeNotificationRule.html#API_DescribeNotificationRule_Errors) returns a `ResourceNotFoundException` instead of `NotFound`. Please update the not found error code accordingly.

aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,193 @@
...
+			{
+				Name:        "status",
+				Description: "The status of the notification rule. Valid statuses are on (sending notifications) or off (not sending notifications).",
+				Type:        proto.ColumnType_STRING,
+				Hydrate:     describeNotificationRule,
+			},
+			{
+				Name:        "event_types",
+				Description: "A list of the event types associated with the notification rule.",
+				Type:        proto.ColumnType_JSON,
```
ParthaI, line 69: Could you please re-arrange the columns based on their type? The JSON columns should be positioned towards the bottom of the table definitions to maintain consistency across the tables.

aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,193 @@
...
+			Tags:    map[string]string{"service": "codestar-notifications", "action": "ListNotificationRules"},
+		},
+		HydrateConfig:     []plugin.HydrateConfig{},
+		GetMatrixItemFunc: SupportedRegionMatrix("codestar-notifications"), // unsure how to get EndpointsID from the package
+		Columns: awsRegionalColumns([]*plugin.Column{
+			{
+				Name:        "arn",
+				Description: "The Amazon Resource Name (ARN) of the notification rule.",
+				Type:        proto.ColumnType_STRING,
+			},
```
ParthaI, line 41: Could you please add the missing column `id` here?
I can see the [ListNotificationRules](https://docs.aws.amazon.com/codestar-notifications/latest/APIReference/API_NotificationRuleSummary.html) API response returns the `Id` property.

aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,193 @@
...
+			KeyColumns: plugin.SingleColumn("arn"),
+			IgnoreConfig: &plugin.IgnoreConfig{
+				ShouldIgnoreErrorFunc: shouldIgnoreErrors([]string{"NotFound", "InvalidParameter"}),
+			},
+			Hydrate: describeNotificationRule,
+			Tags:    map[string]string{"service": "codestar-notifications", "action": "DescribeNotificationRule"},
+		},
+		List: &plugin.ListConfig{
+			Hydrate: listNotificationRules,
+			Tags:    map[string]string{"service": "codestar-notifications", "action": "ListNotificationRules"},
```
ParthaI, line 32: Could we add the optional key qualifiers here to minimize the API response and query time? The API [ListNotificationRules](https://docs.aws.amazon.com/codestar-notifications/latest/APIReference/API_ListNotificationRules.html) supports [Filters](https://docs.aws.amazon.com/codestar-notifications/latest/APIReference/API_ListNotificationRules.html#codestarnotifications-ListNotificationRules-request-Filters) as input parameters.

For reference, please see the code at: [Steampipe Plugin AWS: table_aws_ec2_instance.go](https://github.com/turbot/steampipe-plugin-aws/blob/main/aws/table_aws_ec2_instance.go#L541-L546).

aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+}
+
+//// LIST FUNCTION
+
+func listNotificationRules(ctx context.Context, d *plugin.QueryData, _ *plugin.HydrateData) (interface{}, error) {
+	// Get client
+	svc, err := CodeStarNotificationsClient(ctx, d)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_codestar_notification_rule.listNotificationRules", "get_client_error", err)
```
ParthaI, line 128:
```suggestion
		plugin.Logger(ctx).Error("aws_codestar_notification_rule.listNotificationRules", "connection_error", err)
```

aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	if arn == "" {
+		return nil, nil
+	}
+
+	// Get client
+	svc, err := CodeStarNotificationsClient(ctx, d)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_codestar_notification_rule.describeNotificationRule", "get_client_error", err)
```
ParthaI, line 178:
```suggestion
		plugin.Logger(ctx).Error("aws_codestar_notification_rule.getCodeStarNotificationRule", "connection_error", err)
```

aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+		}
+	}
+
+	return nil, err
+}
+
+//// HYDRATE FUNCTIONS
+
+func describeNotificationRule(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
```
ParthaI, line 162:
```suggestion
func getCodeStarNotificationRule(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
```

Please update the function naming wherever necessary.

aws/table_aws_codestar_notification_rule.go
```diff
@@ -0,0 +1,193 @@
...
+				Hydrate:     describeNotificationRule,
+				Transform:   transform.FromField("Arn").Transform(transform.EnsureStringArray),
+			},
+		}),
+	}
+}
+
+//// LIST FUNCTION
+
+func listNotificationRules(ctx context.Context, d *plugin.QueryData, _ *plugin.HydrateData) (interface{}, error) {
```
ParthaI, line 124:
```suggestion
func listCodeStarNotificationRules(ctx context.Context, d *plugin.QueryData, _ *plugin.HydrateData) (interface{}, error) {
```

Please update the function naming wherever required.

ParthaI (commented):
Hello @gertvv,

I've left a few review comments for you to check. If you encounter any blockers or have concerns, please feel free to let us know.

Additionally, please include the documentation for this table, similar to what we have for the others.

Thank you!

aws/table_aws_codestar_notification_rule.go
gertvv: That sounds good, but after some fiddling I haven't been able provoke an InvalidParameter error, and I'm also not sure where I can handle it. Some pointers for both would be useful.

aws/table_aws_codestar_notification_rule.go
gertvv: I've made this change (and will push it soon), but I think I took the "get_client_error" from another table I used as reference (possibly aws_sns_topic), and am wondering when it is appropriate to use one or the other.

aws/table_aws_codestar_notification_rule.go
gertvv: I've done this and was able to check that I'm still able to filter by the relevant columns, but I haven't been able to verify that this has the intended performance impact.

There are also two possible filters that don't map to any specific columns. Is there anything to be done in this case?

gertvv:
I'm also wondering because it is quite natural to want to do joins on the "targets" - is it possible for the targets to be exposed as a secondary table somehow? The JSON join works but is quite inconvenient.

aws/table_aws_codestar_notification_rule.go
gertvv: I've removed the ignore for InvalidParameter. Using an invalid `where arn = 'abcdef'` clause causes a permissions error instead of this. And nonsense values for the filter parameters do not result in errors.

aws/table_aws_codestar_notification_rule.go
ParthaI: I have noticed that a few tables use the `get_client_error` log statement, while most tables use `connection_error`. To ensure consistency across all tables, I believe it would be better to standardize on using `connection_error`.

aws/table_aws_codestar_notification_rule.go
ParthaI: > There are also two possible filters that don't map to any specific columns. Is there anything to be done in this case?

We can address this by:

- Adding a column to accept the filter parameter:

```go
{
  Name:        "<Column Name>",
  Description: "...",
  Type:        <Column Type>,
  Transform:   transform.FromQual("<Column Name>"),
},
```

- Using this column in the function [buildCodeStarNotificationRulesFilter](https://github.com/turbot/steampipe-plugin-aws/pull/2198/files#diff-d94afafdc25a86b7a1149cfa7f29faba181fa61c7be19d0d01dc0ef7ad1ad903R214-R217) to build the filter parameter.

Note: This column value will be populated when the value is specified as a query parameter in the where clause.

ParthaI:
> I'm also wondering because it is quite natural to want to do joins on the "targets" - is it possible for the targets to be exposed as a secondary table somehow? The JSON join works but is quite inconvenient.
> 
> I added another example query to the documentation that at least makes this more manageable.


- According to our table development standards, we can create a separate table if the API supports pagination for retrieving the column value.
- In our case, if there is a specific API call to get the `targets` only and it supports pagination.
- Otherwise, we will leave them as they are.

The example query looks great to me. However, I have a few minor suggestions:
- Please include the example query in both PostgreSQL and SQLite formats, as we have for other tables.
- Please fix the indentation of the example queries. We can use the tool https://codebeautify.org/sqlformatter

Thank you!

aws/table_aws_codestar_notification_rule.go
gertvv: I've tried to do this but didn't get it to work. The following returns zero rows:
```
select
  name, target_address
from
  aws_codestar_notification_rule
where target_address = 'arn:aws:sns:***' ;
```
Whereas the following returns one result:
```
aws codestar-notifications list-notification-rules \
  --filters Name=TARGET_ADDRESS,Value=arn:aws:sns:***
```
I've pushed the code so would appreciate if you can have a look at it.

gertvv:
>     * According to our table development standards, we can create a separate table if the API supports pagination for retrieving the column value.

Although there is an endpoint for listing targets it doesn't quite do what we'd want here so I'll leave it as is.

> The example query looks great to me. However, I have a few minor suggestions:
> 
>     * Please include the example query in both PostgreSQL and SQLite formats, as we have for other tables.

I couldn't find instructions for building the plugin for SQLite. How do I do that?

ParthaI:
Hmm, It is difficult to build the plugin to work with SQLite in our local.

Can we use online tools like [pg-to-sqlite](https://github.com/hirefrank/pg-to-sqlite) and [RebaseData](https://www.rebasedata.com/convert-postgresql-to-sqlite-online) to format our existing query into SQLite format?

On another note, the changes look good to me. We can merge them into a feature branch in our plugin and then make the remaining changes in that branch.

Thanks!

gertvv:
Awesome, thank you! :+1:

PR #2272, opened 2024-08-19
Add table aws_ec2_load_balancer_listener_rule Closes #2243

### LINKED ISSUE #2243: Add table aws_ec2_load_balancer_listener_rule(s)
**References**
[AWS API](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_DescribeRules.html)
[Go API](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2#Client.DescribeRules)
[listener rules for ALB](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-update-rules.html)

**Use case**

The initial use case I have in mind is to find all targets for all public-facing ALBs, so I know which targets are receiving "public" ingress traffic. This new table (or column) is needed because until now, the only relevant column in `aws_ec2_load_balancer_listener` is `default_actions`, but for anything beyond a trivial use case, we use additional actions or rules to define how to handle requests received by the load balancer, and the Steampipe AWS plugin currently doesn't provide a way to query these actions or rules. This new table or column would provide a way to make queries like this.

**Comments on structure**

The API is called specifying a listener ARN or a list of rule ARNs, and it returns a list of rules. I don't know whether this is better represented as a new table or as a new column in `aws_ec2_load_balancer_listener`.

If this is implemented as a separate table, I don't know whether it's better done as a rule per row (table name `aws_ec2_load_balancer_listener_rule`), or a listener per row that includes a column with all rules for that listener (table name `aws_ec2_load_balancer_listener_rules`).

When one uses the AWS CLI `aws elbv2 describe-rules --listener-arn <arn>`, this command returns all rules for the listener, including the default action which is already in the `default_actions` column of `aws_ec2_load_balancer_listener`.

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
N/A
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```

> select * from aws_ec2_load_balancer_listener_rule where listener_arn = 'arn:aws:elasticloadbalancing:us-east-1:xxxxxxxxxxxx:listener/app/test53333/f7cc8cdc44ff910b/c9418b57592205f0'
+------------------------------------------------------------------------------------------------------------------------------------+----------+------------------------------------------------------------------------------------------>
| arn                                                                                                                                | priority | listener_arn                                                                             >
+------------------------------------------------------------------------------------------------------------------------------------+----------+------------------------------------------------------------------------------------------>
| arn:aws:elasticloadbalancing:us-east-1:xxxxxxxxxxxx:listener-rule/app/test53333/f7cc8cdc44ff910b/c9418b57592205f0/a8fe6d8842838dfa | default  | arn:aws:elasticloadbalancing:us-east-1:xxxxxxxxxxxx:listener/app/test53333/f7cc8cdc44ff91>
+------------------------------------------------------------------------------------------------------------------------------------+----------+------------------------------------------------------------------------------------------


>  select jsonb_pretty(actions) from aws_ec2_load_balancer_listener_rule where arn = 'arn:aws:elasticloadbalancing:us-east-1:xxxxxxxxxxxx:listener-rule/app/test53333/f7cc8cdc44ff910b/c9418b57592205f0/a8fe6d8842838dfa'
+---------------------------------------------------------------------------------------------------------------------------------+
| jsonb_pretty                                                                                                                    |
+---------------------------------------------------------------------------------------------------------------------------------+
| [                                                                                                                               |
|     {                                                                                                                           |
|         "Type": "forward",                                                                                                      |
|         "Order": null,                                                                                                          |
|         "ForwardConfig": {                                                                                                      |
|             "TargetGroups": [                                                                                                   |
|                 {                                                                                                               |
|                     "Weight": 1,                                                                                                |
|                     "TargetGroupArn": "arn:aws:elasticloadbalancing:us-east-1:xxxxxxxxxxxx:targetgroup/test53/c27339fc272ca4cf" |
|                 }                                                                                                               |
|             ],                                                                                                                  |
|             "TargetGroupStickinessConfig": {                                                                                    |
|                 "Enabled": false,                                                                                               |
|                 "DurationSeconds": null                                                                                         |
|             }                                                                                                                   |
|         },                                                                                                                      |
|         "RedirectConfig": null,                                                                                                 |
|         "TargetGroupArn": "arn:aws:elasticloadbalancing:us-east-1:xxxxxxxxxxxx:targetgroup/test53/c27339fc272ca4cf",            |
|         "FixedResponseConfig": null,                                                                                            |
|         "AuthenticateOidcConfig": null,                                                                                         |
|         "AuthenticateCognitoConfig": null                                                                                       |
|     }                                                                                                                           |
| ]                                                                                                                               |
+---------------------------------------------------------------------------------------------------------------------------------+
```
</details>

### DIFF
--- a/aws/canonical_policy_test.go
+++ b/aws/canonical_policy_test.go
@@ -266,7 +266,7 @@ func TestConvertPolicy(t *testing.T) {
 func prettyPrint(src interface{}) {
 	pretty, err := json.MarshalIndent(src, "", "  ")
 	if err != nil {
-		log.Fatalf(err.Error())
+		log.Fatal(err.Error())
 	}
 	fmt.Printf("\n %s\n", string(pretty))
 
--- a/aws/plugin.go
+++ b/aws/plugin.go
@@ -220,7 +220,7 @@ func Plugin(ctx context.Context) *plugin.Plugin {
 			"aws_ec2_classic_load_balancer":                                tableAwsEc2ClassicLoadBalancer(ctx),
 			"aws_ec2_client_vpn_endpoint":                                  tableAwsEC2ClientVPNEndpoint(ctx),
 			"aws_ec2_gateway_load_balancer":                                tableAwsEc2GatewayLoadBalancer(ctx),
-			"aws_ec2_load_balancer_listener_rule": tableAwsEc2ApplicationLoadBalancerListenerRule(ctx),
+			"aws_ec2_load_balancer_listener_rule":                          tableAwsEc2ApplicationLoadBalancerListenerRule(ctx),
 			"aws_ec2_instance":                                             tableAwsEc2Instance(ctx),
 			"aws_ec2_instance_availability":                                tableAwsInstanceAvailability(ctx),
 			"aws_ec2_instance_metric_cpu_utilization":                      tableAwsEc2InstanceMetricCpuUtilization(ctx),
--- a/aws/plugin.go
+++ b/aws/plugin.go
@@ -220,6 +220,7 @@ func Plugin(ctx context.Context) *plugin.Plugin {
 			"aws_ec2_classic_load_balancer":                                tableAwsEc2ClassicLoadBalancer(ctx),
 			"aws_ec2_client_vpn_endpoint":                                  tableAwsEC2ClientVPNEndpoint(ctx),
 			"aws_ec2_gateway_load_balancer":                                tableAwsEc2GatewayLoadBalancer(ctx),
+			"aws_ec2_load_balancer_listener_rule": tableAwsEc2ApplicationLoadBalancerListenerRule(ctx),
 			"aws_ec2_instance":                                             tableAwsEc2Instance(ctx),
 			"aws_ec2_instance_availability":                                tableAwsInstanceAvailability(ctx),
 			"aws_ec2_instance_metric_cpu_utilization":                      tableAwsEc2InstanceMetricCpuUtilization(ctx),
--- a/aws/table_aws_account.go
+++ b/aws/table_aws_account.go
@@ -183,7 +183,7 @@ func getOrganizationDetails(ctx context.Context, d *plugin.QueryData, _ *plugin.
 func accountDataToTitle(ctx context.Context, d *transform.TransformData) (interface{}, error) {
 	accountInfo := d.HydrateItem.(*accountData)
 
-	if accountInfo.Aliases != nil && len(accountInfo.Aliases) > 0 {
+	if len(accountInfo.Aliases) > 0 {
 		return accountInfo.Aliases[0], nil
 	}
 
--- a/aws/table_aws_appautoscaling_target.go
+++ b/aws/table_aws_appautoscaling_target.go
@@ -195,7 +195,7 @@ func getAwsApplicationAutoScalingTarget(ctx context.Context, d *plugin.QueryData
 		plugin.Logger(ctx).Error("aws_appautoscaling_target.getAwsApplicationAutoScalingTarget", "api_error", err)
 		return nil, err
 	}
-	if op.ScalableTargets != nil && len(op.ScalableTargets) > 0 {
+	if len(op.ScalableTargets) > 0 {
 		return op.ScalableTargets[0], nil
 	}
 
--- a/aws/table_aws_cloudtrail_trail.go
+++ b/aws/table_aws_cloudtrail_trail.go
@@ -319,7 +319,7 @@ func getCloudtrailTrail(ctx context.Context, d *plugin.QueryData, _ *plugin.Hydr
 		return nil, err
 	}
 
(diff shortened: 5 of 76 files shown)


misraved (commented):
@ParthaI please take a look at the review comments. Thanks!!

dkewley:
I initiated the request for this table in issue #2243.

The docs in this code currently say:

> You **_must_** specify `arn` or `listener_arn` in a `where` clause in order to use this table.

From this I understand that the table will only contain at most a single row if a rule `arn` is specified, or possibly multiple rows which contain all the rules for a single `listener_arn`. Is that correct?

If I want to derive a table that contains all the listener rules for all the listeners associated with a single ALB, or all the listeners in a Steampipe connection (e.g. as in table `aws_ec2_load_balancer_listener`), how would I derive such a table? Can you include that as an example query in the table docs?

Would it instead be reasonable to include in this initial build of this table an ability for the caller to request that the table include all listener rules for the specified Steampipe connection? (I think I'm wording this right; let me know if you want clarification.) I think this would ease the use case I mentioned in #2243.

ParthaI:
Hi @dkewley,

> From this I understand that the table will only contain at most a single row if a rule arn is specified, or possibly multiple rows which contain all the rules for a single listener_arn. Is that correct?

Yes, correct.

> If I want to derive a table that contains all the listener rules for all the listeners associated with a single ALB, or all the listeners in a Steampipe connection (e.g. as in table aws_ec2_load_balancer_listener), how would I derive such a table? Can you include that as an example query in the table docs?

I have updated the table doc with more example queries that will provide the all the listener rules for all the listeners by joining the table `aws_ec2_load_balancer_listener`.

```sql
 select
  r.arn,
  r.listener_arn,
  l.load_balancer_arn,
  l.protocol as listener_protocol,
  l.ssl_policy,
  r.priority,
  r.is_default,
  r.actions,
  r.conditions
from
  aws_ec2_load_balancer_listener_rule as r
  join aws_ec2_load_balancer_listener as l on r.listener_arn = l.arn;
```

> Would it instead be reasonable to include in this initial build of this table an ability for the caller to request that the table include all listener rules for the specified Steampipe connection? (I think I'm wording this right; let me know if you want clarification.) I think this would ease the use case I mentioned in https://github.com/turbot/steampipe-plugin-aws/issues/2243.

Yes, it is possible to configure a parent hydrate for the `aws_ec2_load_balancer_listener_rule` table.

However, there is a limitation: Steampipe does not currently support _PARENT HYDRATE CHAINING_ by design. This means that since the `aws_ec2_load_balancer_listener` table already has a [parent hydrate configured](https://github.com/turbot/steampipe-plugin-aws/blob/main/aws/table_aws_ec2_load_balancer_listener.go#L33), we cannot use `aws_ec2_load_balancer_listener` as the parent of `aws_ec2_load_balancer_listener_rule`.

That said, we could achieve the desired functionality by redesigning the `aws_ec2_load_balancer_listener_rule` table as follows:
- Use a parent hydrate to list all the load balancers.
- Make API calls to list the listeners for each load balancer.
- Iterate over the listeners to make API calls for listing their rules.

However, please note that this design may significantly increase the query time for the `aws_ec2_load_balancer_listener_rule` table. From a query performance perspective, this approach may not be ideal. 

Alternatively, with the existing design, you can achieve the same result by joining the `aws_ec2_load_balancer_listener_rule` and `aws_ec2_load_balancer_listener` tables.

On the other hand, we can join the `aws_ec2_application_load_balancer` and `aws_ec2_network_load_balancer` tables with the `aws_ec2_load_balancer_listener_rule` and `aws_ec2_load_balancer_listener` tables. This approach can help optimize query timing and performance.

E:g. Get listener rules for application load balancers

```sql
 select
  r.arn,
  r.listener_arn,
  l.load_balancer_arn,
  l.protocol as listener_protocol,
  l.ssl_policy,
  a.canonical_hosted_zone_id,
  a.dns_name,
  a.ip_address_type,
  r.priority,
  r.is_default,
  r.actions,
  r.conditions
from
  aws_ec2_load_balancer_listener_rule as r
  join aws_ec2_load_balancer_listener as l on r.listener_arn = l.arn
  join aws_ec2_application_load_balancer as a on l.load_balancer_arn = a.arn;
```

Please let us know if the above discussion finds you well.

Thanks!

dkewley:
Hi @ParthaI,

Thank you so much for your thorough and thoughtful response. It helps me understand the things you're balancing in this table design a bit better.

I think with the added query examples, it probably makes sense to leave this table with faster performance. Then users such as myself can choose to build on those example queries as needed, accepting a use-case-specific increase in overall query time in order to build up larger result sets / tables.

Also, thanks to you and the Turbot team for being quite responsive to new-table requests such as this. It really helps end-users such as myself to lean even more into relying on Steampipe for our operational reporting needs.

misraved (approved):

PR #2332, opened 2024-10-31
Bug Fix: Wrong Comparison Operator for Number and Date Filter in `aws_inspector2_finding`

dbermuehler:
This PR fixes a nasty bug which made the `aws_inspector2_finding` table returned no results if the SQL query contains the `<=` or `>=` operator for number or date columns.

To explain the bug take this query:

```sql
select
  title,
  epss_score
from
  aws_inspector2_finding
where
  epss_score >= 0.7
```

In this case, the call to the Inspector2 findings API should contain the NumberFilter for the EPSS score field with the field `LowerInclusive` and the value 0.7. However, in the code the SQL operator `>=` was mapped to the NumberFilter field `UpperInclusive`. This leads to the API response with exactly the opposite results we actually wanted to retrieve. When the list function returned the result list to Steampipe, the SQL query operator was applied to the result set, which then filtered out all the results.

This bug affected all columns for which the NumberFilter and DateFilter was applied.

# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
```
</details>

### DIFF
--- a/aws/table_aws_inspector2_finding.go
+++ b/aws/table_aws_inspector2_finding.go
@@ -850,9 +850,9 @@ func buildNumberFilters(d *plugin.QueryData, filter *types.FilterCriteria) {
 				val := aws.Float64(q.Value.GetDoubleValue())
 				var f types.NumberFilter
 				switch q.Operator {
-				case "<=":
-					f = types.NumberFilter{LowerInclusive: val}
 				case ">=":
+					f = types.NumberFilter{LowerInclusive: val}
+				case "<=":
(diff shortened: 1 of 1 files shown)


misraved:
Thanks for catching the bug @dbermuehler 👍!

Could you please add a few example queries demonstrating the error and how the change in code fixes the bug?

dbermuehler:
The query in the PR description should be able to show case the different results you get with and without the bug fix. Or would you like to see something in particular?

misraved (approved):

PR #2408, opened 2025-02-10
Fix typo integration test failure

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
```
</details>


aws/table_aws_kinesis_consumer.go
```diff
@@ -141,6 +143,12 @@ func listKinesisConsumers(ctx context.Context, d *plugin.QueryData, h *plugin.Hy
 
 		output, err := paginator.NextPage(ctx)
 		if err != nil {
+			var ae smithy.APIError
+			if errors.As(err, &ae) {
+				if ae.ErrorCode() == "ResourceNotFoundException" {
```
cbruno10, line 148: @ParthaI Why do we check for the error here instead of having `ResourceNotFoundException` in the `ListConfig`'s `IgnoreConfig`?

ParthaI: @cbruno10, we use parent hdrate(`ParentHydrate: listStreams,`) for this table.
In the case of parent hydrate the ignore config seems to not work for the child table. So we need to handle it manually.
Steampipe SDK issue ref: https://github.com/turbot/steampipe-plugin-sdk/issues/544

cbruno10: Thanks @ParthaI , that makes sense and the comment helps!

cbruno10 (approved):

PR #2410, opened 2025-02-11
Resolved `InvalidParameterException` in `aws_ecs_service` table when listing tags for older services. Closes #2388

### LINKED ISSUE #2388: Tags cannot be fetched for `aws_ecs_service` table
**Describe the bug**

> ERROR:  rpc error: code = Unknown desc = my_aws_account: operation error ECS: ListTagsForResource, https response error StatusCode: 400, RequestID: c5b155da-7eb3-42d7-a94b-d7f3cf2e32d7, InvalidParameterException: Long arn format must be used for tagging operations

> ERROR:  rpc error: code = Unknown desc = my_aws_account: table 'aws_ecs_service' column 'tags' requires hydrate data from getEcsServiceTags, which failed with error operation error ECS: ListTagsForResource, https response error StatusCode: 400, RequestID: 076ed52f-8f0e-43b9-af89-3728995bb52b, InvalidParameterException: Long arn format must be used for tagging operations.

Steampipe v1.0.2

`hub.steampipe.io/plugins/turbot/aws@latest | 1.5.0`

**To reproduce**
```sql
select 
  account_id,
  region,
  arn,
  tags -- Fetching tags causes the issue
from
  aws_ecs_service;
```

**Expected behavior**
Tags for `aws_ecs_service` should get returned like for other tables.

**Additional context**
None

ParthaI:
# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
> select service_name, tags_src, tags from aws_ecs_service;
+-----------------+-------------------------------------------------------------+-----------------------------+
| service_name    | tags_src                                                    | tags                        |
+-----------------+-------------------------------------------------------------+-----------------------------+
| test-53-service | [{"Key":"help","Value":"me"},{"Key":"foo1","Value":"bar1"}] | {"foo1":"bar1","help":"me"} |
+-----------------+-------------------------------------------------------------+-----------------------------+
```
</details>

### DIFF
--- a/aws/table_aws_ecs_service.go
+++ b/aws/table_aws_ecs_service.go
@@ -2,13 +2,12 @@ package aws
 
 import (
 	"context"
-	"errors"
+	"strings"
 	"sync"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ecs"
 	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
-	"github.com/aws/smithy-go"
 
 	ecsv1 "github.com/aws/aws-sdk-go/service/ecs"
 
@@ -223,6 +222,11 @@ func tableAwsEcsService(_ context.Context) *plugin.Table {
 	}
 }
 
+type ServiceInfo struct {
+	ClusterName *string
+	types.Service
+}
+
 //// LIST FUNCTION
 
 func listEcsServices(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
@@ -297,7 +301,7 @@ func listEcsServices(ctx context.Context, d *plugin.QueryData, h *plugin.Hydrate
 
 	for result := range serviceCh {
 		for _, service := range result.Services {
-			d.StreamListItem(ctx, service)
+			d.StreamListItem(ctx, ServiceInfo{cluster.ClusterName, service})
 
 			// Context may get cancelled due to manual cancellation or if the limit has been reached
 			if d.RowsRemaining(ctx) == 0 {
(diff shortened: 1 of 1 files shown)


misraved:
@ParthaI has the `arn` format changed completely?

ParthaI:
@Misraved,

Yes, the ARN format has changed, and currently, AWS supports both formats, as mentioned in [this documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-ids).

We might need to upgrade the resource, but it's unclear how long AWS will continue supporting the old format.

Tasks launched by an Amazon ECS service can only receive the new ARN and resource ID format if the service was created on or after November 16, 2018, and the user who created the service has opted in to the new format for tasks.

The timeline for the opt-in and opt-out periods for the new Amazon Resource Name (ARN) and resource ID format for Amazon ECS resources ended on April 1, 2021. By default, all accounts are opted in to the new format. All new resources created receive the new format, and you can no longer opt out.

misraved:
@ParthaI can we update the plugin code to use the new ARN format to make the `ListTagsForResource` API call?

thomasklemm:
@ParthaI Managed to install it and run the plugin from this branch. Now I'm getting the following error.

```
ERROR:  rpc error: code = Internal desc = aws_production: rpc error: code = Internal desc = hydrate function getEcsServiceTags failed with panic runtime error: invalid memory address or nil pointer dereference
```

Query:
```sql
select 
  account_id,
  region,
  arn,
  tags -- Fetching tags causes the issue
from
  aws_ecs_service;
```

It's again pretty hard to pinpoint it to the specific record causing it. Does the error message already help you?

ParthaI:
Ah, my apologies for the oversight. I have now pushed a fix, and it should work as expected.  

The root cause of the issue was that the **list API call for the cluster did not return the `ClusterName`**. To retrieve the cluster name, we need to make a **Get API call** instead.  

Could you please try again after pulling the latest changes from the branch?  

**Note:** _**It would be helpful to verify whether the cluster name is being returned for older clusters before running the query to fetch the tags for the services.**_  

Thank you!

thomasklemm:
@ParthaI I think it's a good step further, but I get another error: 

> ERROR:  rpc error: code = Unknown desc = aws_1234567890: operation error ECS: ListTagsForResource, https response error StatusCode: 400, RequestID: 84644e43-0be9-4075-8aa1-c4c1ecc70bf0, InvalidParameterException: The specified service does not have the longArn format. Only services created after opting in will have the new format.

In terms of cluster names, `select * from aws_ecs_cluster` (not `_service`) is able to return all the cluster names for clusters of any age with this PR (I can't say right now how it behaved before but can roll it back and and check). The `aws_ecs_service` table doesn't have a cluster name column as far as I see, only the `cluster_arn`, that one seems to get returned properly.

Query:
```
select 
  service.account_id,
  service.region,
  service.arn,
  service.cluster_arn,
  cluster.cluster_name as cluster_name, -- All cluster names look fine here
  cluster.tags as cluster_tags
  -- service.tags as service_tags -- This causes the error
from
  aws_ecs_service service
left join aws_ecs_cluster cluster
  on cluster.cluster_arn = service.cluster_arn;
```

ParthaI:
Hmm, the error indicates that we cannot make the API call by updating the ARN format for the older service.

Could you please confirm whether the AWS CLI command successfully retrieves the tag details for the older service? If so, could you kindly share the CLI command along with the specific ARN format used for the older service? This would help us update the table code accordingly to match the correct format.

ParthaI:
Furthermore, according to the AWS documentation, retrieving tag details for older services is not supported. To tag a service that follows the older ARN format, you must migrate it to the **long ARN format**.  

**Documentation Reference:** [AWS ECS API - TagResource](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TagResource.html)

misraved (approved):

PR #2481, opened 2025-04-30
Add aws_macie2_finding table

misraved:
# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
```
</details>


aws/table_aws_macie2_finding.go
```diff
@@ -0,0 +1,316 @@
...
+		List: &plugin.ListConfig{
+			Hydrate: listMacie2Findings,
+			Tags:    map[string]string{"service": "macie2", "action": "ListFindings"},
+			KeyColumns: []*plugin.KeyColumn{
+				{Name: "finding_type", Require: plugin.Optional, Operators: []string{"=", "<>"}},
+				{Name: "severity", Require: plugin.Optional, Operators: []string{"=", "<>"}},
+				{Name: "status", Require: plugin.Optional, Operators: []string{"=", "<>"}},
+			},
+		},
+		GetMatrixItemFunc: SupportedRegionMatrix(macie2v1.EndpointsID),
```
ParthaI, line 41: Since we're in the process of removing AWS SDK V1 support from the plugin, can we use the endpoint ID from `endpoint_service_ids_gen.go` instead?

aws/table_aws_macie2_finding.go
```diff
@@ -0,0 +1,316 @@
...
+		Columns: awsRegionalColumns([]*plugin.Column{
+			{
+				Name:        "id",
+				Description: "The unique identifier for the finding.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "finding_arn",
```
ParthaI, line 49:
```suggestion
				Name:        "arn",
```

aws/table_aws_macie2_finding.go
```diff
@@ -0,0 +1,316 @@
...
+				Description: "The Amazon Resource Name (ARN) of the finding.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("FindingArn"),
+			},
+			{
+				Name:        "finding_type",
```
ParthaI, line 55: Was this intentional? Shouldn't the column name align with the field name in the API response?

aws/table_aws_macie2_finding.go
```diff
@@ -0,0 +1,316 @@
...
+			},
+			{
+				Name:        "created_at",
+				Description: "The date and time, in UTC and extended ISO 8601 format, when the finding was created.",
+				Type:        proto.ColumnType_TIMESTAMP,
+			},
+			{
+				Name:        "updated_at",
+				Description: "The date and time, in UTC and extended ISO 8601 format, when the finding was last updated.",
+				Type:        proto.ColumnType_TIMESTAMP,
```
ParthaI, line 78: Do you think it would be better if we handle potential null values? What does the API return if the finding has never been updated?

aws/table_aws_macie2_finding.go
```diff
@@ -0,0 +1,316 @@
...
+				Name:        "source_account_id",
+				Description: "The AWS account ID for the account that owns the finding.",
+				Type:        proto.ColumnType_STRING,
+			},
+
+			// Steampipe standard columns
+			{
+				Name:        "title",
+				Description: resourceInterfaceDescription("title"),
+				Type:        proto.ColumnType_STRING,
```
ParthaI, line 135: Shouldn’t we add a transform function here? Does the API actually return a field named `title`?

aws/table_aws_macie2_finding.go
```diff
@@ -0,0 +1,316 @@
...
+			if strings.Contains(err.Error(), "Macie is not enabled.") {
+				return nil, nil
+			}
+			plugin.Logger(ctx).Error("aws_macie2_finding.listMacie2Findings", "api_error", err)
+			return nil, err
+		}
+
+		for _, findingId := range output.FindingIds {
+			// Get finding details
+			finding, err := getMacie2Finding(ctx, d, &plugin.HydrateData{
```
ParthaI, line 209: Here we're making a nested API call based on the number of results from the list call. Do you think it would be better to keep the list function minimal—streaming only the finding ID like we do in other tables—and use a hydrate function (get call) to populate the additional column values?

ParthaI (commented):
Hi @misraved, I have provided an initial review comment; please take a look. Thanks!

aws/table_aws_macie2_finding.go
ParthaI: Removed the column from the table since AWS does not treat Macie Finding as a resource type. Reference: [AWS Macie Resource Types](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/AWS_Macie.html).
Additionally, the List and Get operations do not return an ARN value in their response.

aws/table_aws_macie2_finding.go
ParthaI: Updated the column name to `type`.

aws/table_aws_macie2_finding.go
ParthaI: Added the appropriate transform function.

aws/table_aws_macie2_finding.go
ParthaI: Yes, the API response has field named `Title`.

aws/table_aws_macie2_finding.go
ParthaI: Removed the nested API call from list function.

ParthaI (approved):

Issue #2514, opened 2025-06-04
The issue where the `aws_msk_cluster` table returns a maximum of only three records.


insummersnow:
**Describe the bug**
The `aws_msk_cluster` table returns a maximum of only 3 records. This issue is caused by the SDK's `GetBootstrapBrokers` API, which returns only up to 3 broker endpoints.

**Steampipe version (`steampipe -v`)**
Example: v0.23.3

**Plugin version (`steampipe plugin list`)**
Example: v1.14.1

**To reproduce**
Run a query against the `aws_msk_cluster` table in an environment where the number of brokers exceeds 3.
Observe that only 3 entries are returned regardless of the actual number of brokers in the cluster.

**Expected behavior**
All broker endpoints in the MSK cluster should be returned, not just the first three.

**Additional context**
The fix involves analyzing the broker endpoint pattern, which follows the format b-n.blahblah.kafka.region.amazonaws.com:port.
Using GetBootstrapBrokers, we extract the blahblah and port segments from the returned endpoint.
Then, we use DescribeClusterV2 to determine the total number of brokers.
Based on this count, we construct the full list of broker endpoint URLs through iteration.
A PR will be submitted shortly with this fix.

![Image](https://github.com/user-attachments/assets/26d6e435-423a-4b71-af83-64927f23f3aa)

![Image](https://github.com/user-attachments/assets/ffa5302c-b06a-4b4d-beeb-88f8f64bd3dd)

ParthaI:
Hi @insummersnow 👋

As discussed in the PR ([#2515](https://github.com/turbot/steampipe-plugin-aws/pull/2515)), the API currently returns a maximum of only three records, which appears to be a limitation on the AWS side.

To address this, we've raised support requests with the AWS CLI and SDK teams:

* AWS CLI: [aws/aws-cli#9557](https://github.com/aws/aws-cli/issues/9557)
* AWS SDK for Go v2: [aws/aws-sdk-go-v2#3116](https://github.com/aws/aws-sdk-go-v2/issues/3116)

In the meantime, we're closing this issue for now. We'll revisit and reopen it once the underlying limitation is resolved by AWS.

Thanks so much for your understanding!

PR #2515, opened 2025-06-04
bugfix : https://github.com/turbot/steampipe-plugin-aws/issues/2514

insummersnow:
https://github.com/turbot/steampipe-plugin-aws/issues/2514

# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example query results
steampipe-plugin-aws_issues_2514_screenshot1
steampipe-plugin-aws_issues_2514_screenshot2

<details>
  <summary>Results</summary>

```
```
</details>

### DIFF
--- a/aws/table_aws_msk_cluster.go
+++ b/aws/table_aws_msk_cluster.go
@@ -2,6 +2,8 @@ package aws
 
 import (
 	"context"
+	"strings"
+	"fmt"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/kafka"
@@ -101,12 +103,14 @@ func tableAwsMSKCluster(_ context.Context) *plugin.Table {
 				Description: "A string containing one or more hostname:port pairs of Kafka brokers suitable for use with Apache Kafka clients.",
 				Type:        proto.ColumnType_STRING,
 				Hydrate:     getKafkaClusterBootstrapBrokers,
+				Transform:   transform.FromField("BootstrapBrokerString"),
 			},
 			{
 				Name:        "bootstrap_broker_string_tls",
 				Description: "A string containing one or more hostname:port pairs of Kafka brokers suitable for TLS authentication.",
 				Type:        proto.ColumnType_STRING,
 				Hydrate:     getKafkaClusterBootstrapBrokers,
+				Transform:   transform.FromField("BootstrapBrokerStringTls"),
 			},
 			{
 				Name:        "provisioned",
@@ -342,15 +346,71 @@ func getKafkaClusterBootstrapBrokers(ctx context.Context, d *plugin.QueryData, h
 		return nil, nil
 	}
 
-	params := &kafka.GetBootstrapBrokersInput{
+	describeParams := &kafka.DescribeClusterV2Input{
 		ClusterArn: &clusterArn,
 	}
-
-	op, err := svc.GetBootstrapBrokers(ctx, params)
+	describeResp, err := svc.DescribeClusterV2(ctx, describeParams)
 	if err != nil {
-		logger.Error("aws_msk_cluster.getKafkaClusterBootstrapBrokers", "api_error", err)
+		logger.Error("aws_msk_cluster.getKafkaClusterBootstrapBrokers", "describe_cluster_api_error", err)
+		return nil, err
+	}
+	brokerCount := int(*describeResp.ClusterInfo.Provisioned.NumberOfBrokerNodes)
+
+	bootstrapParams := &kafka.GetBootstrapBrokersInput{
+		ClusterArn: &clusterArn,
+	}
+	bootstrapResp, err := svc,GetBootstrapBrokers(ctx, bootstrapParams)
+	if err != nil {
+		logger.Error("aws_msk_cluster.getKafkaClusterBootstrapBrokers", "get_bootstrap_api_error", err)
 		return nil, err
 	}
 
-	return op, nil
+	samplingBroker := ""
+	if bootstrapResp.BootstrapBrokerString != nil {
+		parts := strings.Split(*bootstrapResp.BootstrapBrokerString, ",")
+		if len(parts) > 0 {
+			samplingBroker = parts[0]
+		}
+	}
+	if samplingBroker == "" {
+		return bootstrapResp, nil
+	}
+
+	var plainPort = "9092"
+	var tlsPort = "9094"
+	if bootstrapResp.BootstrapBrokerString != nil {
+		if parts := strings.Split(*bootstrapResp.BootstrapBrokerString, ":"); len(parts) == 2 {
+			plainPort = parts[1]
+		}
+	}
+	if bootstrapResp.BootstrapBrokerStringTls != nil {
+		if parts := strings.Split(*bootstrapResp.BootstrapBrokerStringTls, ":"); len(parts) == 2 {
+			tlsPort = parts[1]
+		}
+	}
+
+	hostOnly := strings.Split(samplingBroker, ":")[0]
+	dotParts := strings.Split(hostOnly, ".", 2)
+	if len(dotParts) != 2 {
+		logger.Warn("aws_msk_cluster.getKafkaClusterBootstrapBrokers", "invalid_hostname", samplingBroker)
+		return nil, nil
+	}
+	suffix := dotParts[1]
+
+	plainBrokers := make([]string, brokerCount)
+	tlsBrokers := make([]string, brokerCount)
+	for i := 1; i <= brokerCount; i++ {
+		plainBrokers[i-1] = fmt.Sprintf("b-%d.%s:%s", i, suffix, plainPort)
+		tlsBrokers[i-1] = fmt.Sprintf("b-%d.%s:%s", i, suffix, tlsPort)
+	}
+
+	result := struct {
+		BootstrapBrokerString string
+		BootstrapBrokerStringTls string
+	}{
+		BootstrapBrokerString: strings.Join(plainBrokers, ","),
+		BootstrapBrokerStringTls: strings.Join(tlsBrokers, ","),
+	}
+
+	return result, nil
 }


insummersnow:
Sure, I'll take a look at the linting issue and push an update shortly. Thanks for the heads-up!

misraved (approved):

aws/table_aws_msk_cluster.go
```diff
@@ -342,15 +346,71 @@ func getKafkaClusterBootstrapBrokers(ctx context.Context, d *plugin.QueryData, h
...
+	if bootstrapResp.BootstrapBrokerString != nil {
+		if parts := strings.Split(*bootstrapResp.BootstrapBrokerString, ":"); len(parts) == 2 {
+			plainPort = parts[1]
+		}
+	}
+	if bootstrapResp.BootstrapBrokerStringTls != nil {
+		if parts := strings.Split(*bootstrapResp.BootstrapBrokerStringTls, ":"); len(parts) == 2 {
+			tlsPort = parts[1]
+		}
+	}
```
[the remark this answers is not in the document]
misraved: @insummersnow could you please take a look at this review comment?

insummersnow:
@misraved Please give me a review~!!

aws/table_aws_msk_cluster.go
```diff
@@ -342,15 +346,75 @@ func getKafkaClusterBootstrapBrokers(ctx context.Context, d *plugin.QueryData, h
 		return nil, nil
 	}
 
-	params := &kafka.GetBootstrapBrokersInput{
+	describeParams := &kafka.DescribeClusterV2Input{
 		ClusterArn: &clusterArn,
 	}
+	describeResp, err := svc.DescribeClusterV2(ctx, describeParams)
```
ParthaI: Do we need to make the `DescribeClusterV2` API call again here? It seems that the list API call already provides the necessary details.

aws/table_aws_msk_cluster.go
```diff
@@ -342,15 +346,75 @@ func getKafkaClusterBootstrapBrokers(ctx context.Context, d *plugin.QueryData, h
...
+	if err != nil {
+		logger.Error("aws_msk_cluster.getKafkaClusterBootstrapBrokers", "describe_cluster_api_error", err)
+		return nil, err
+	}
+	brokerCount := int(*describeResp.ClusterInfo.Provisioned.NumberOfBrokerNodes)
```
ParthaI: I noticed that the list call is already returning the `NumberOfBrokerNodes` value.

 Would it be possible to retrieve it using something like the following?

 ```go
 cluster := h.Item.(types.Cluster)
 numberOfBrokerNodes := *cluster.Provisioned.NumberOfBrokerNodes
 ```

aws/table_aws_msk_cluster.go
```diff
@@ -342,15 +346,75 @@ func getKafkaClusterBootstrapBrokers(ctx context.Context, d *plugin.QueryData, h
...
+		if len(parts) > 0 {
+			samplingBroker = parts[0]
+		}
+	}
+	if samplingBroker == "" {
+		return bootstrapResp, nil
+	}
+
+	var plainPort = parsePort(bootstrapResp.BootstrapBrokerString, "9092")
+	var tlsPort = parsePort(bootstrapResp.BootstrapBrokerString, "9094")
```
ParthaI, line 371: It seems we're using a hard-coded value here. Will this value be consistent across all MSK clusters? In my case, the broker list looks like: `b-2.lowcostmsk.n3w692.c5.kafka.us-east-1.amazonaws.com:9098,b-3.lowcostmsk.n3w692.c5.kafka.us-east-1.amazonaws.com:9098,b-1.lowcostmsk.n3w692.c5.kafka.us-east-1.amazonaws.com:9098`. Could you please update it to use a dynamic value instead?

aws/table_aws_msk_cluster.go
```diff
@@ -342,15 +346,75 @@ func getKafkaClusterBootstrapBrokers(ctx context.Context, d *plugin.QueryData, h
...
+	}
+	brokerDomainSuffix := dotParts[1]
+
+	plainBrokers := make([]string, brokerCount)
+	tlsBrokers := make([]string, brokerCount)
+
+	// Broker URL pattern (e.g. b-1.blahblah.kafka.ap-northeast-2.amazonaws.com:9092)
+	// Port 9092 is used for plaintext connections, 9094 for TLS within VPC
+	// Public TLS endpoints may use port 9194
+	// https://docs.aws.amazon.com/cli/latest/reference/kafka/get-bootstrap-brokers.html
```
ParthaI: It would be better to use AWS CLI v2, as all of our APIs are relevant to it and we also use AWS SDK v2.

aws/table_aws_msk_cluster.go
```diff
@@ -101,12 +103,14 @@ func tableAwsMSKCluster(_ context.Context) *plugin.Table {
 				Description: "A string containing one or more hostname:port pairs of Kafka brokers suitable for use with Apache Kafka clients.",
 				Type:        proto.ColumnType_STRING,
 				Hydrate:     getKafkaClusterBootstrapBrokers,
+				Transform:   transform.FromField("BootstrapBrokerString"),
```
ParthaI: Do we need to explicitly include the transform function here? Isn't the default transform function (`DefaultTransform: transform.FromCamel()`) defined at the plugin level sufficient? I believe it should help populate the column value correctly.

aws/table_aws_msk_cluster.go
```diff
@@ -342,15 +346,75 @@ func getKafkaClusterBootstrapBrokers(ctx context.Context, d *plugin.QueryData, h
...
+		tlsBrokers[i-1] = fmt.Sprintf("b-%d.%s:%s", i, brokerDomainSuffix, tlsPort)
+	}
+
+	result := struct {
+		BootstrapBrokerString string
+		BootstrapBrokerStringTls string
+	}{
+		BootstrapBrokerString: strings.Join(plainBrokers, ","),
+		BootstrapBrokerStringTls: strings.Join(tlsBrokers, ","),
+	}
```
ParthaI, line 399: It looks like we're modifying the API response here, which we generally try to avoid. I agree that the API isn't returning the expected result, but in this case, wouldn't it be more appropriate to raise a support request with the AWS CLI or SDK team to address the issue on their side?

@misraved, What is your thoughts on this?

aws/table_aws_msk_cluster.go
```diff
@@ -101,12 +103,14 @@ func tableAwsMSKCluster(_ context.Context) *plugin.Table {
...
 				Type:        proto.ColumnType_STRING,
 				Hydrate:     getKafkaClusterBootstrapBrokers,
+				Transform:   transform.FromField("BootstrapBrokerString"),
 			},
 			{
 				Name:        "bootstrap_broker_string_tls",
 				Description: "A string containing one or more hostname:port pairs of Kafka brokers suitable for TLS authentication.",
 				Type:        proto.ColumnType_STRING,
 				Hydrate:     getKafkaClusterBootstrapBrokers,
+				Transform:   transform.FromField("BootstrapBrokerStringTls"),
```
ParthaI: Do we need to explicitly include the transform function here? Isn't the default transform function (`DefaultTransform: transform.FromCamel()`) defined at the plugin level sufficient?

ParthaI (commented):
Hi @insummersnow, I've left a few review comments—could you please take a look when you have a moment?

aws/table_aws_msk_cluster.go
insummersnow: Hi @Parthal, apologies for the delayed response.

I'm parsing the port for use, and if it can't be parsed, it defaults to a hardcoded port. You can refer to the function below.
func parsePort(brokerString *string, defaultPort string) string {
	if brokerString != nil {
		if parts := strings.Split(*brokerString, ":"); len(parts) == 2 {
			return parts[1]
		}
	}
	return defaultPort
}

aws/table_aws_msk_cluster.go
insummersnow: Yes, I agree~
I'll go ahead and update it.

aws/table_aws_msk_cluster.go
insummersnow: Yes, I agree~
I'll go ahead and update it.

aws/table_aws_msk_cluster.go
insummersnow: Yes, I agree~
I'll go ahead and update it.

aws/table_aws_msk_cluster.go
insummersnow: I completely agree with your point, which is why I’ve submitted a request for the CLI to return all brokers by default:
https://github.com/aws/aws-cli/issues/9557

If the CLI is updated to return the full list of brokers, I expect the SDK will reflect that as well. I’ll verify this once the feature request is accepted.

If the request is accepted and the CLI is updated accordingly so that the SDK also returns all brokers, I’ll revert to the original logic.
Until then, I think it makes sense to handle it this way to avoid functional issues. What do you think?

aws/table_aws_msk_cluster.go
insummersnow: Yes, I agree~
I'll go ahead and update it. :)

aws/table_aws_msk_cluster.go
insummersnow: Yes, I agree~
I'll go ahead and update it. :)

aws/table_aws_msk_cluster.go
ParthaI: Perfect, thanks @insummersnow! It would be helpful to [attach this link for context](https://github.com/aws/aws-cli/issues/9557) to set the expectation. I’d also recommend raising a request in the SDK as well.

aws/table_aws_msk_cluster.go
ParthaI: Yes, I understand your logic and can confirm it’s working as expected. However, as per Steampipe development standards, we avoid making manual changes to the API response. That’s why I suggested raising the issue on the SDK/CLI side.

ParthaI:
Hi @insummersnow, please let me know once you’ve addressed all the review comments — I’ll take another look at the PR then. Thanks!

aws/table_aws_msk_cluster.go
insummersnow: I completely agree with your point.
Thank you for highlighting this — it’s an important limitation that deserves to be addressed.

The related issues have been opened here:
AWS CLI: https://github.com/aws/aws-cli/issues/9557
AWS SDK for Go v2: https://github.com/aws/aws-sdk-go-v2/issues/3116

aws/table_aws_msk_cluster.go
insummersnow: I completely agree with your point.
Thank you for highlighting this — it’s an important limitation that deserves to be addressed.

The related issues have been opened here:
AWS CLI: https://github.com/aws/aws-cli/issues/9557
AWS SDK for Go v2: https://github.com/aws/aws-sdk-go-v2/issues/3116

misraved:
@insummersnow we are closing this PR for now since we are waiting for an update from the AWS CLI and SDK team - https://github.com/aws/aws-cli/issues/9557

Once we have the response, feel free to reopen the PR, thanks 👍 !!

PR #2561, opened 2025-07-17
Add default rate-limit configuration for AWS Lambda tables

misraved:
# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
```
</details>


aws/plugin.go
```diff
@@ -175,6 +175,38 @@ func Plugin(ctx context.Context) *plugin.Plugin {
...
+				Scope:      []string{"connection", "region", "service", "action"},
+				Where:      "service = 'lambda' and action = 'ListFunctions'",
+			},
+			{
+				// GetFunctionUrlConfig is part of the control plane APIs sharing 15 rps limit
+				Name:       "aws_lambda_get_function_url_config",
+				FillRate:   10,
+				BucketSize: 10,
+				Scope:      []string{"connection", "region", "service", "action"},
+				Where:      "service = 'lambda' and action = 'GetFunctionUrlConfig'",
```
ParthaI: @misraved, regarding the rate-limit configuration for `aws_lambda_list_functions` and `aws_lambda_get_function_url_config`, we currently have a limit of `10` requests per second. Does this imply that I can issue up to 10 `ListFunctions` and 10 `GetFunctionUrlConfig` API calls in parallel?

If that’s not the case, instead of maintaining two separate rate-limiter configurations, would it be more appropriate to define a single configuration such as:

```go
{
    // All other control plane APIs (including ListFunctions) share a limit of 15 requests per second
    Name:       "aws_lambda_list_functions",
    FillRate:   10,
    BucketSize: 10,
    Scope:      []string{"connection", "region", "service", "action"},
    Where:      "service = 'lambda' and action in ('ListFunctions', 'GetFunctionUrlConfig')",
},
```

Could you please confirm whether combining them under a single configuration is the correct approach?

ParthaI (approved):

PR #2636, opened 2025-09-09
Availability Zone retrieval does not respect regions in aws.spc

kazvictor:
Availability zones are currently listed from all regions `ParentHydrate: listAwsRegions`. This is an issue when service control policies are set restricting certain regions. For example:
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Condition": {
        "StringNotEquals": {
          "aws:RequestedRegion": [
            "us-east-1",
            "us-east-2",
            "us-west-1",
            "us-west-2"
          ]
        }
      },
      "Resource": "*",
      "Effect": "Deny"
    }
  ]
}
```

Having a policy like the above will cause queries like: `select * from myconfig.aws_availability_zone` to fail even if in the `aws.scp` you restrict regions: `regions = ["us-west-1", "us-west-2", "us-east-1", "us-east-2"]`

The fix is to instead use the `GetMatrixItemFunc: SupportedRegionMatrix(ec2v1.EndpointsID)` and `EC2Client(ctx, d)` like most other aws tables.

### Integration test logs
<details>
  <summary>Logs</summary>

```
N/A, no integration tests for availability zones
```
</details>

### DIFF
--- a/aws/table_aws_availability_zone.go
+++ b/aws/table_aws_availability_zone.go
@@ -6,7 +6,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
-
+	ec2v1 "github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/turbot/steampipe-plugin-sdk/v5/grpc/proto"
 	"github.com/turbot/steampipe-plugin-sdk/v5/plugin"
 	"github.com/turbot/steampipe-plugin-sdk/v5/plugin/transform"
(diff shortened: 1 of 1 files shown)


ParthaI:
Hello @kazvictor, I’ve reviewed the code changes and they look great overall. However, I have a couple of technical questions:

* Do these changes correctly handle scenarios where an SCP restricts access to certain regions?
* Since both `ParentHydrate: listAwsRegions` and `GetMatrixItemFunc: SupportedRegionMatrix(ec2v1.EndpointsID)` internally invoke the [DescribeRegions](https://github.com/turbot/steampipe-plugin-aws/blob/main/aws/multi_region.go#L488) API through [listRawAwsRegions](https://github.com/turbot/steampipe-plugin-aws/blob/main/aws/multi_region.go#L461), I would expect an error to be raised if SCP restrictions block API calls for those regions. Could you confirm if that is indeed the case?

Thanks in advance for clarifying!

ParthaI:
Furthermore, I was able to reproduce the behavior by configuring the connection and SCPs as described in this PR.

From what I recall, the `aws_region` and `aws_availability_zone` tables were originally designed to return all regions/availability zones for `opted-in` regions. For this reason, we did not use `GetMatrixItemFunc`, since that would restrict results to only the regions configured in the connection settings.

Introducing iteration of `DescribeAvailabilityZones` over the regions defined in the connection would therefore be a breaking change. For example, with the current `main` branch:

* No SCPs are applied in the organization.
* My Steampipe connection is configured to a single region: `regions = ["us-west-1"]`.
* Running the query `SELECT * FROM aws_availability_zone` currently returns all available zones across all opted-in regions.
* With the changes in this PR, the same query would return zones only for `us-west-1`, which alters the existing behavior.

As an alternative, if you want to avoid errors while maintaining broader coverage, you can configure either:

* `ignore_error_codes = ["UnauthorizedOperation"]`, or
* `ignore_error_messages = [".*an explicit deny in a service control policy.*"]`.

This will bypass the error and return results as expected.

By the way, I identified a known bug in the table where these ignore configurations were not being honored. To address this, I’ve raised [a PR](https://github.com/turbot/steampipe-plugin-aws/pull/2640). Please try it out in that branch with the `ignore_error_codes` or `ignore_error_messages` configuration in your `.spc` file and let me know if you’re able to get the expected results.

Thanks!

PR #2702, opened 2026-02-04
Fetch all tags associated with each EC2 instance

### LINKED ISSUE #2701: aws_ec2_instance table fetches only the first 10 tags on each EC2 instance
**Describe the bug**
AWS Steampipe only fetches the first 10 tags on each EC2 instance because it uses the describe instance API call to retrieve instance tags. This API call does not fetch all tags associated with the instance and instead only returns the first 10. To solve this, AWS Steampipe should make individual API calls to fetch EC2 instance tags.

v2.3.4

v1.29.0

**To reproduce**
Create an EC2 instance and add 11 or more tags to the instance. When Steampipe scrapes this instance, the tags column will contain only 10 of the tags.

**Expected behavior**
When Steampipe scrapes instances, it should retrieve all of the tags associated with each EC2 instance

nwernink-squid:
Resolves https://github.com/turbot/steampipe-plugin-aws/issues/2701 by fetching all tags associated with each EC2 instance.

# Integration test logs
<details>
  <summary>Logs</summary>

```
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
```
</details>

### DIFF
--- a/aws/table_aws_ec2_instance.go
+++ b/aws/table_aws_ec2_instance.go
@@ -86,6 +86,10 @@ func tableAwsEc2Instance(_ context.Context) *plugin.Table {
 				Func: getInstanceStatus,
 				Tags: map[string]string{"service": "ec2", "action": "DescribeInstanceStatus"},
 			},
+			{
+				Func: getEc2InstanceTags,
+				Tags: map[string]string{"service": "ec2", "action": "DescribeTags"},
+			},
 		},
 		GetMatrixItemFunc: SupportedRegionMatrix(AWS_EC2_SERVICE_ID),
 		Columns: awsRegionalColumns([]*plugin.Column{
@@ -494,7 +498,8 @@ func tableAwsEc2Instance(_ context.Context) *plugin.Table {
 				Name:        "tags_src",
 				Description: "A list of tags assigned to the instance.",
 				Type:        proto.ColumnType_JSON,
-				Transform:   transform.FromField("Tags"),
+				Hydrate:     getEc2InstanceTags,
+				Transform:   transform.FromValue(),
 			},
 
 			// Steampipe standard columns
@@ -508,6 +513,7 @@ func tableAwsEc2Instance(_ context.Context) *plugin.Table {
 				Name:        "tags",
 				Description: resourceInterfaceDescription("tags"),
 				Type:        proto.ColumnType_JSON,
+				Hydrate:     getEc2InstanceTags,
 				Transform:   transform.From(getEc2InstanceTurbotTags),
 			},
 			{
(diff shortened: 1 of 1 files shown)


nwernink-squid:
> @nwernink-squid thank you so much for raising the PR, could you please take a look at the review comments generated by copilot?

Hey @misraved, yes sorry for the delay

misraved (changes requested):
## Code Review

### BLOCKER 1: `title` column still reads from limited tags

The `getEc2InstanceTurbotTitle` transform function type-asserts `d.HydrateItem.(types.Instance)` and reads `data.Tags` from the original `DescribeInstances` response, which is still limited to 10 tags. If the "Name" tag falls outside the first 10, `title` will be incorrect.

The `title` column needs to either:
- Specify `Hydrate: getEc2InstanceTags` and update the transform to handle `[]types.TagDescription`, or
- Have the transform check both the instance tags and the hydrate results

### BLOCKER 2: `tags_src` schema change is breaking

Previously `tags_src` returned `[]types.Tag` (fields: `Key`, `Value`). Now it returns `[]types.TagDescription` which includes additional `ResourceId` and `ResourceType` fields in every element. This is a **breaking change** for downstream consumers (queries, dashboards, mods that parse `tags_src`).

Fix by converting back to the original schema in the hydrate function or a transform:

```go
var result []types.Tag
for _, td := range tags {
    result = append(result, types.Tag{Key: td.Key, Value: td.Value})
}
return result, nil
```

### BLOCKER 3: N+1 API calls without rate limit protection

`getEc2InstanceTags` makes one `DescribeTags` call **per instance**. For accounts with thousands of instances, this creates a significant N+1 problem and risks API throttling.

The `HydrateConfig` should set `MaxConcurrency` to limit parallel calls, e.g.:

```go
{
    Func:           getEc2InstanceTags,
    MaxConcurrency: 10,
    Tags:           map[string]string{"service": "ec2", "action": "DescribeTags"},
},
```

---

### SUGGESTION: Add error ignore config for terminated instances

If an instance is terminated between the list call and the hydrate call, `DescribeTags` could return an error. Consider adding:

```go
ShouldIgnoreErrorFunc: shouldIgnoreErrors([]string{"InvalidInstanceID.NotFound"}),
```

to the hydrate config. (Verify whether `DescribeTags` with a resource-id filter returns this error or just empty results.)

### SUGGESTION: Paginator is unnecessary

AWS EC2 instances have a maximum of 50 tags. A single `DescribeTags` call with the default max results will always return all tags in one page. A direct API call would be simpler:

```go
output, err := svc.DescribeTags(ctx, params)
if err != nil {
    return nil, err
}
return output.Tags, nil
```

### SUGGESTION: PR body has placeholder text

Both the integration test logs and example query results sections contain placeholder text. Actual evidence should be included — especially showing an instance with >10 tags returning complete results.

---

### NIT: Nil pointer safety

`*instance.InstanceId` is dereferenced without a nil check. While `InstanceId` should never be nil for a valid instance, a guard would be defensive.

---

### Positives

- Addresses a real bug (#2701) — the 10-tag limit from `DescribeInstances` is a genuine issue
- Correct hydrate signature and registration pattern
- Proper rate limiting tags on the hydrate config
- Correct paginator pattern (even if unnecessary here)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

cbruno10:
@nwernink-squid I'm closing this PR as I wasn't able to replicate locally, and also responded in https://github.com/turbot/steampipe-plugin-aws/issues/2701#issuecomment-4578952808, but if you are still experiencing this error and can share your queries + results, feel free to re-open. Thanks!

PR #2716, opened 2026-03-16
Recompile plugin with steampipe-plugin-sdk v5.14.0 and Go version 1.26.0

misraved:


cbruno10:
Plugin has been bumped to a newer SDK version along with Go 1.26 already

PR #2545, opened 2025-07-03
Add ignore error config to aws_rds_db_instance and aws_rds_pending_maintenance_action tables fixes #2543

misraved:
# Integration test logs
<details>
  <summary>Logs</summary>

```
Add passing integration test logs here
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```
Before the fix:

> select
  a.db_instance_identifier,
  b.action,
  a.status,
  b.opt_in_status,
  b.forced_apply_date,
  b.current_apply_date,
  b.auto_applied_after_date
from
  aws_rds_db_instance as a
  join aws_rds_pending_maintenance_action as b on b.resource_identifier = a.arn;

Error: aws_a: operation error RDS: DescribePendingMaintenanceActions, https response error StatusCode: 400, RequestID: b6c9c600-35e3-4969-9340-8a20782ef111, api error InvalidParameterValue: The parameter Filter: resource-id is not a valid identifier. Identifiers must begin with a letter; must contain only ASCII letters, digits, and hyphens; and must not end with a hyphen or contain two consecutive hyphens.
	aws_b: operation error RDS: DescribePendingMaintenanceActions, https response error StatusCode: 400, RequestID: e6d6e682-b692-496b-994a-9bb04b016111, api error InvalidParameterValue: The parameter Filter: resource-id is not a valid identifier. Identifiers must begin with a letter; must contain only ASCII letters, digits, and hyphens; and must not end with a hyphen or contain two consecutive hyphens. (SQLSTATE HV000)

+------------------------+---------------+-----------+---------------+-------------------+--------------------+-------------------------+
| db_instance_identifier | action        | status    | opt_in_status | forced_apply_date | current_apply_date | auto_applied_after_date |
+------------------------+---------------+-----------+---------------+-------------------+--------------------+-------------------------+
| test           | system-update | available | <null>        | <null>            | <null>             | <null>                  |
+------------------------+---------------+-----------+---------------+-------------------+--------------------+-------------------------+

After the fix:

> select
  a.db_instance_identifier,
  b.action,
  a.status,
  b.opt_in_status,
  b.forced_apply_date,
  b.current_apply_date,
  b.auto_applied_after_date
from
  aws_rds_db_instance as a
  join aws_rds_pending_maintenance_action as b on b.resource_identifier = a.arn;
+------------------------+---------------+-----------+---------------+-------------------+--------------------+-------------------------+
[...]

PR #2740, opened 2026-04-29
fix(s3): skip hydrate calls for buckets outside name/region qualifiers

### LINKED ISSUE #2737: Optimize aws_s3_bucket: Skip hydrate calls for buckets outside requested region qualifier
When querying `aws_s3_bucket` with a `WHERE region = 'ap-southeast-2'` qualifier, Steampipe still runs **all 14 hydrate API calls on every bucket across all regions** before filtering. This causes severe timeouts for accounts with a large number of buckets.

### Our scenario:
- Account has **272 S3 buckets** spread across `ap-south-1` (182), `ap-southeast-1` (66), `us-east-1` (20), `me-south-1` (2), `eu-west-1` (2)
- **0 buckets** exist in the requested region (`ap-southeast-2`)
- Despite this, Steampipe fires **~3,800 API calls** (272 buckets × 14 hydrate calls each)
- 2 buckets in `me-south-1` caused TCP-level `i/o timeout` on all hydrate calls, each retrying 9 times with exponential backoff
- Query got **stuck for 12+ minutes** at 270/272 rows before hitting `DeadlineExceeded`
- The final result was **0 rows** — all that work for nothing

### The flow today:
1. `ListBuckets` (global API) → returns all 272 buckets (no region info)
2. `HeadBucket` on each → determines region
3. **All 14 hydrate calls** on ALL 272 buckets (ACL, Policy, Tags, Encryption, Versioning, Lifecycle, Logging, Website, ObjectLock, OwnershipControls, EventNotification, PublicAccessBlock, PolicyStatus, ARN)
4. Region filter applied **after** all hydrate calls complete
5. Returns only matching rows

### Hydrate calls per bucket:
| # | Function | AWS API |
|---|---|---|
| 1 | getBucketRegion | HeadBucket |
| 2 | getBucketACL | GetBucketAcl |
| 3 | getBucketPolicy | GetBucketPolicy |
| 4 | getBucketIsPublic | GetBucketPolicyStatus |
| 5 | getBucketTagging | GetBucketTagging |
| 6 | getBucketEncryption | GetBucketEncryption |
| 7 | getBucketVersioning | GetBucketVersioning |
| 8 | getBucketLifecycle | GetLifecycleConfiguration |
| 9 | getBucketWebsite | GetBucketWebsite |
| 10 | getBucketLogging | GetBucketLogging |
| 11 | getObjectLockConfiguration | GetObjectLockConfiguration |
| 12 | getS3BucketObjectOwnershipControl | GetBucketOwnershipControls |
| 13 | getS3BucketEventNotificationConfigurations | GetBucketNotificationConfiguration |
| 14 | getBucketPublicAccessBlock | GetPublicAccessBlock |

## Describe the solution you'd like

After `HeadBucket` determines a bucket's region, **skip all remaining hydrate calls if the bucket's region does not match the `WHERE region = ?` qualifier**. 

### Proposed flow:
1. `ListBuckets` → returns all buckets
2. `HeadBucket` on each → determines region
3. **If region doesn't match qualifier → skip hydrate calls, don't stream this row**
4. Only run the 13 remaining hydrate calls on buckets that match the region
5. Return matching rows

In our case, this would reduce **~3,800 API calls to ~272** (just HeadBucket), and the query would return instantly instead of timing out.

## Context

I'm aware that PR #2536 attempted this optimization but was reverted in #2545 due to `InvalidToken` errors. This request is to re-attempt this optimization with a fix for the credential scoping issue.

## Query that triggers this

```sql
SELECT account_id, acl, akas, arn, block_public_acls, block_public_policy, 
       bucket_policy_is_public, creation_date, event_notification_configuration, 
       ignore_public_acls, lifecycle_rules, logging, name, object_lock_configuration, 
       object_ownership_controls, partition, policy, policy_std, region, 
       restrict_public_buckets, server_side_encryption_configuration, tags, tags_src, 
       title, versioning_enabled, versioning_mfa_delete, website_configuration 
FROM aws_s3_bucket 
WHERE account_id = '742465441725' AND region = 'ap-southeast-2'
```

## Error

Error: rpc error: code = DeadlineExceeded desc = aws: exceeded allowed timeout (SQLSTATE HV000)

From plugin logs, the final timeout was caused by:
```
GetBucketNotificationConfiguration on bi-global-reports-01.s3.me-south-1.amazonaws.com
→ dial tcp 52.95.172.40:443: i/o timeout
→ exceeded maximum number of attempts, 9
A bucket in me-south-1 (not even the requested region) caused the entire query to fail.

```
## Environment
Steampipe v2.3.5
steampipe-plugin-aws (latest)
Account with 272 S3 buckets across multiple regions
Connection configured with single region

rmnjaat:
## Problem

Querying `aws_s3_bucket` with `WHERE region = 'x'` or `WHERE name = 'x'`
triggered all 14 hydrate API calls on **every bucket globally** before
any filtering occurred.

For an account with 276 buckets and `WHERE region = 'ap-southeast-2'`
(zero matching buckets), this caused ~3,800 API calls and query timeouts
exceeding 12 minutes despite returning zero results.

Fixes #2737

## Root Cause

- `ListBuckets` always returns all buckets globally (unavoidable)
- `getBucketRegion` (HeadBucket) ran for every bucket regardless of filters
- All 13 downstream hydrates ran for every bucket regardless of filters
- Region/name filtering only happened at the SQL engine level, after all API calls completed

## Solution

Two early-exit gates inserted at the earliest possible stage:

### Gate 1 — Name filter (inside `listS3Buckets`, before `StreamListItem`)

`name` is available directly from `ListBuckets`. Buckets not matching a
`name = '...'` or `name IN (...)` qualifier are skipped before
`d.StreamListItem`, so they **never enter the hydrate pipeline**.

Saves: **HeadBucket + 13 downstream API calls** per non-matching bucket.

### Gate 2 — Region filter (inside `doGetBucketRegion`, after HeadBucket)

After HeadBucket resolves the actual bucket region, buckets outside the
requested region return `nil, nil` from `getBucketRegion`. All 13
dependent hydrate functions have a nil guard at the top and exit
immediately without making any API calls.

Saves: **13 API calls** per non-matching bucket (HeadBucket is unavoidable
as region is unknown before it runs).

Both gates handle `=` and `IN` operators via `d.Quals` loops. Other
operators fall back to full hydration. The region cache path also applies
the qualifier check so warm-cache queries are equally optimised.

## Performance Impact

| Scenario (276 total buckets) | Before | After |
|---|---|---|
| `WHERE region = 'ap-southeast-2'` (0 match) | ~3,864 calls, 12+ min | ~276 calls, ~1.3s |
| `WHERE region = 'us-east-1'` (21 match) | ~3,864 calls | ~276 + 273 = 549 calls |
| `WHERE name = 'my-bucket'` (1 match) | ~3,864 calls | ~15 calls, ~0.2s |
| No filter | ~3,864 calls | ~3,864 calls (unchanged) |

## Notes

- Does **not** create region-scoped S3 clients (avoids the `InvalidToken` issue that caused PR #2536 to be reverted in #2545)
- `name` and `region` declared as `OptionalColumns` in `ListConfig` solely to make `d.Quals` accessible in hydrate functions — `listS3Buckets` still fetches all buckets globally
- No behaviour change for queries without `name` or `region` filters

# Integration test logs

<details>
<summary>Logs</summary>

```
$ go test ./aws/... -v
--- PASS: TestConvertPolicy/32 (0.00s)
--- PASS: TestConvertPolicy/33 (0.00s)
--- PASS: TestConvertPolicy/34 (0.00s)
PASS
ok      github.com/turbot/steampipe-plugin-aws/aws      0.830s
```

</details>

# Example query results

<details>
<summary>Results</summary>

```
-- Test 1: name = filter (1 match from 276 buckets)
> select name, region, versioning_enabled, bucket_policy_is_public, tags
  from aws_s3_bucket
  where name = 'my-app-assets-prod';

+--------------------+------------+--------------------+-------------------------+----------------------------------------------+
| name               | region     | versioning_enabled | bucket_policy_is_public | tags                                         |
+--------------------+------------+--------------------+-------------------------+----------------------------------------------+
| my-app-assets-prod | ap-south-1 | false              | false                   | {"Environment":"Prod","Team":"platform"}     |
+--------------------+------------+--------------------+-------------------------+----------------------------------------------+


-- Test 2: name IN filter (2 matches)
> select name, region, versioning_enabled
  from aws_s3_bucket
  where name in ('my-app-assets-prod', 'my-app-logs-dev');

+--------------------+------------+--------------------+
| name               | region     | versioning_enabled |
+--------------------+------------+--------------------+
| my-app-assets-prod | ap-south-1 | false              |
| my-app-logs-dev    | ap-south-1 | false              |
+--------------------+------------+--------------------+


-- Test 3: region = filter (21 matches)
> select name, region
  from aws_s3_bucket
  where region = 'us-east-1'
  order by name;

+------------------------------+-----------+
| name                         | region    |
+------------------------------+-----------+
| my-alb-access-logs-01        | us-east-1 |
| my-app-cdn-logs              | us-east-1 |
| my-app-cloudfront-01         | us-east-1 |
| my-app-ses-logs              | us-east-1 |
| my-backup-bucket-01          | us-east-1 |
| my-billing-reports           | us-east-1 |
| my-cdn-assets                | us-east-1 |
| my-cloudtrail-logs           | us-east-1 |
| my-cost-reports-main         | us-east-1 |
| my-cur-bucket-123456789012   | us-east-1 |
| my-db-backups                | us-east-1 |
| my-deploy-artifacts          | us-east-1 |
| my-elb-logs-prod             | us-east-1 |
| my-lambda-artifacts          | us-east-1 |
| my-ops-cur-bucket            | us-east-1 |
| my-root-activity-alerts      | us-east-1 |
| my-s3-access-logs-03         | us-east-1 |
| my-serveraccess-logs         | us-east-1 |
| my-static-assets             | us-east-1 |
| my-temp-logs-use1            | us-east-1 |
| my-waf-logs-cloudfront       | us-east-1 |
+------------------------------+-----------+


-- Test 4: region = filter with 0 matches
> select name, region
  from aws_s3_bucket
  where region = 'ap-southeast-2';

+------+--------+
| name | region |
+------+--------+
+------+--------+


-- Test 5: region IN filter (multiple regions, 23 matches)
> select name, region
  from aws_s3_bucket
  where region in ('us-east-1', 'eu-west-1')
  order by region, name;

+------------------------------+-----------+
| name                         | region    |
+------------------------------+-----------+
| my-eu-app-uploads            | eu-west-1 |
| my-eu-temp-logs              | eu-west-1 |
| my-alb-access-logs-01        | us-east-1 |
| my-app-cdn-logs              | us-east-1 |
| my-app-cloudfront-01         | us-east-1 |
| my-app-ses-logs              | us-east-1 |
| my-backup-bucket-01          | us-east-1 |
| my-billing-reports           | us-east-1 |
| my-cdn-assets                | us-east-1 |
| my-cloudtrail-logs           | us-east-1 |
| my-cost-reports-main         | us-east-1 |
| my-cur-bucket-123456789012   | us-east-1 |
| my-db-backups                | us-east-1 |
| my-deploy-artifacts          | us-east-1 |
| my-elb-logs-prod             | us-east-1 |
| my-lambda-artifacts          | us-east-1 |
| my-ops-cur-bucket            | us-east-1 |
| my-root-activity-alerts      | us-east-1 |
| my-s3-access-logs-03         | us-east-1 |
| my-serveraccess-logs         | us-east-1 |
| my-static-assets             | us-east-1 |
| my-temp-logs-use1            | us-east-1 |
| my-waf-logs-cloudfront       | us-east-1 |
+------------------------------+-----------+


-- Test 6: no filter — baseline unchanged (276 total buckets)
> select count(*) as total from aws_s3_bucket;

+-------+
| total |
+-------+
| 276   |
+-------+
```

</details>

### DIFF
--- a/aws/table_aws_s3_bucket.go
+++ b/aws/table_aws_s3_bucket.go
@@ -20,8 +20,9 @@ func tableAwsS3Bucket(_ context.Context) *plugin.Table {
 		Name:        "aws_s3_bucket",
 		Description: "AWS S3 Bucket",
 		List: &plugin.ListConfig{
-			Hydrate: listS3Buckets,
-			Tags:    map[string]string{"service": "s3", "action": "ListBucket"},
+			Hydrate:    listS3Buckets,
+			KeyColumns: plugin.OptionalColumns([]string{"name", "region"}),
+			Tags:       map[string]string{"service": "s3", "action": "ListBucket"},
 		},
 
 		// Note: No Get for S3 buckets, since it must list all the buckets
@@ -316,6 +317,12 @@ func listS3Buckets(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateDa
 	}
 
 	for _, bucket := range bucketsResult.Buckets {
+		// Filter by name before streaming — bucket never enters the hydrate pipeline,
+		// saving HeadBucket + all 13 downstream API calls for non-matching buckets.
+		if !bucketNameMatchesQuals(d, *bucket.Name) {
+			plugin.Logger(ctx).Debug("aws_s3_bucket.listS3Buckets", "skipping_bucket", *bucket.Name)
+			continue
+		}
 		d.StreamListItem(ctx, bucket)
 		// Context may get cancelled due to manual cancellation or if the limit has been reached
 		if d.RowsRemaining(ctx) == 0 {
@@ -336,7 +343,11 @@ func doGetBucketRegion(ctx context.Context, d *plugin.QueryData, h *plugin.Hydra
 	commonColumnData := c.(*awsCommonColumnData)
 	cacheKey := "getBucketRegion/" + commonColumnData.Partition + "/" + bucket
 	if cachedData, ok := d.ConnectionManager.Cache.Get(cacheKey); ok {
-		return cachedData.(string), nil
+		bucketRegion := cachedData.(string)
+		if !bucketRegionMatchesQuals(d, bucketRegion) {
+			return "", nil
+		}
+		return bucketRegion, nil
 	}
 
 	// Create client in default region
@@ -362,18 +373,75 @@ func doGetBucketRegion(ctx context.Context, d *plugin.QueryData, h *plugin.Hydra
 	plugin.Logger(ctx).Debug("aws_s3_bucket.doGetBucketRegion", "bucket", bucket, "region", bucketRegion)
 
 	d.ConnectionManager.Cache.Set(cacheKey, bucketRegion)
+
+	if !bucketRegionMatchesQuals(d, bucketRegion) {
+		return "", nil
+	}
 	return bucketRegion, nil
+}
+
+// bucketRegionMatchesQuals returns true when bucketRegion satisfies the region quals in the query
+// (e.g. region = 'us-east-1' or region IN ('us-east-1', 'ap-south-1')), or when no region qual
+// is present. Only = operator quals are evaluated; other operators fall back to full hydration.
+//
+// NOTE: region is declared as an OptionalColumns key column solely to make d.Quals["region"]
+// reachable inside hydrate functions. The list function itself does not filter by region at the
+// API level — S3 buckets are always listed globally.
+func bucketRegionMatchesQuals(d *plugin.QueryData, bucketRegion string) bool {
+	regionQuals, ok := d.Quals["region"]
+	if !ok || regionQuals == nil {
+		return true // no region filter — always hydrate
+	}
+	for _, q := range regionQuals.Quals {
+		if q.Operator == "=" && q.Value.GetStringValue() == bucketRegion {
+			return true
+		}
+	}
+	return false
 }
 
 func getBucketRegion(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
 	bucketName := h.Item.(types.Bucket).Name
 
-	return doGetBucketRegion(ctx, d, h, *bucketName)
+	// name is known from ListBuckets — check it before HeadBucket to save all 14 hydrate calls
+	// per non-matching bucket (region filter can only save 13, after HeadBucket already ran).
+	if !bucketNameMatchesQuals(d, *bucketName) {
+		return nil, nil
+	}
+
+	region, err := doGetBucketRegion(ctx, d, h, *bucketName)
+	if err != nil {
+		return nil, err
+	}
+	if region == "" {
+		return nil, nil
+	}
+	return region, nil
+}
+
+// bucketNameMatchesQuals returns true when bucketName satisfies the name quals in the query
+// (e.g. name = 'my-bucket' or name IN ('a', 'b')), or when no name qual is present.
+// Only = operator quals are evaluated; other operators fall back to full hydration.
+func bucketNameMatchesQuals(d *plugin.QueryData, bucketName string) bool {
+	nameQuals, ok := d.Quals["name"]
+	if !ok || nameQuals == nil {
+		return true // no name filter — always hydrate
+	}
+	for _, q := range nameQuals.Quals {
+		if q.Operator == "=" && q.Value.GetStringValue() == bucketName {
+			return true
+		}
+	}
+	return false
 }
 
 func getS3BucketEventNotificationConfigurations(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
 	bucketName := h.Item.(types.Bucket).Name
-	bucketRegion := h.HydrateResults["getBucketRegion"].(string)
+	bucketRegionRaw := h.HydrateResults["getBucketRegion"]
+	if bucketRegionRaw == nil {
+		return nil, nil
+	}
+	bucketRegion := bucketRegionRaw.(string)
 
 	// Create client
 	svc, err := S3Client(ctx, d, bucketRegion)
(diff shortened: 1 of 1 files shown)


cbruno10:
Hi @rmnjaat , thanks for this, and apologies for the slow review. We confirmed that today a query like `where name = 'my-bucket'` still lists every bucket and runs all the per-bucket hydrate calls before filtering, so there's a lot of wasted API calls here.

Before we can merge though, there are a couple of issues I noticed:

1. When there are multiple `in` conditions in the `where` clause, 0 rows are returned. For example:

```sql
select
  name,
  region
from
  aws_s3_bucket
where
  name in ('bucket-a', 'bucket-b')
  and region in ('us-east-1', 'us-west-2');
```

The SDK doesn't split the list quals (it only does that when a single key column has a list value), so `GetStringValue()` returns `""` for both and no bucket matches. The query comes back empty instead of returning the matching buckets.

2. The non-`=` operator handling doesn't fall back the way the description says. The matcher functions return `false` (drop the row) when there's no `=` match, so a query like `where name like 'prod%'` would return nothing rather than falling back to full hydration. It happens to be safe right now only because `OptionalColumns` registers just the `=` operator, but that's fragile.

To keep this simpler and avoid both issues, I think we could just scope the changes down to `name`. `region` can't really be pushed down anyway since `ListBuckets` is global and we have to resolve each bucket's region with a `HeadBucket` call regardless. Limiting to `name` removes the multi-column `in` failure entirely and still captures the win of skipping the ~15 per-bucket detail calls for non-matching buckets. The `nil` guards you added in the dependent hydrates are good.

Could you also share before and after results for the queries below, run against an account with a few buckets across a couple of regions, so we can confirm there are no regressions? Both the row counts and the bucket names should match what `main` returns:

```sql
-- baseline, no quals
select name, region from aws_s3_bucket order by name;

-- single name
select name, region, acl from aws_s3_bucket where name = '<bucket-1>';

-- single-column in
select name, region from aws_s3_bucket where name in ('<bucket-1>', '<bucket-2>');

-- multi-column in (the empty-result case above)
select name, region from aws_s3_bucket
where name in ('<bucket-1>', '<bucket-2>') and region in ('us-east-1', 'us-west-2');

-- non-equals operator
select name, region from aws_s3_bucket where name like '<prefix>%';

-- region only
select name, region from aws_s3_bucket where region = 'us-east-1';
```

Thanks!

rmnjaat:
@cbruno10  Thanks for  the feedback I will push the changes ASAP.

cbruno10:
@rmnjaat We haven't heard back from you in a while, so we're closing the PR, but feel free to re-open. Thanks!

PR #2770, opened 2026-06-04
feat: add aws_application_signals_service_level_objective table

leongzhiyong:
# Integration test logs
Not applicable due to lack of Terraform support for Application Signals service level objectives.

# Example query results
<details>
  <summary>Results</summary>

```
> select * from aws_application_signals_service_level_objective
+----------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------+----------------------+-------------+-----------------+----------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------+----------------------+--------------------+-------------------------------------------------+--------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------+-------------------+-------------------------------------------------------------------------------------------------------------+-----------------+--------------------------+-----------------+-----------------------+------------------------+---------------------------+---------------------------------------------------------------------+----------+-------------------+----------+----------------+-----------------+---------------+------------------------------+--------------------------+-------------------------+--------------------------+-----------+----------------+------------+--------------------+----------------------------------------------------------------+----------------------------------------------------------------+
| arn                                                                                                      | burn_rate_configurations                                                                    | created_time         | description | evaluation_type | goal                                                                                                     | id                                                                                                       | last_updated_time    | metric_source_type | name                                            | operation_name           | request_based_sli                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                | sli    | tags              | dependency_config                                                                                           | dependency_type | dependency_resource_type | dependency_name | dependency_identifier | dependency_environment | dependency_operation_name | key_attributes                                                      | slo_type | slo_resource_type | slo_name | slo_identifier | slo_environment | metric_source | metric_source_key_attributes | metric_source_attributes | include_linked_accounts | slo_owner_aws_account_id | partition | region         | account_id | sp_connection_name | sp_ctx                                                         | _ctx                                                           |
+----------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------+----------------------+-------------+-----------------+----------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------+----------------------+--------------------+-------------------------------------------------+--------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------+-------------------+-------------------------------------------------------------------------------------------------------------+-----------------+--------------------------+-----------------+-----------------------+------------------------+---------------------------+---------------------------------------------------------------------+----------+-------------------+----------+----------------+-----------------+---------------+------------------------------+--------------------------+-------------------------+--------------------------+-----------+----------------+------------+--------------------+----------------------------------------------------------------+----------------------------------------------------------------+
| arn:aws:application-signals:ap-southeast-1:<omitted>:slo/is-this-metric-source                           | <null>                                                                                      | 2026-05-07T06:20:11Z | <null>      | RequestBased    | {"AttainmentGoal":99.9,"Interval":{"Value":{"Duration":1,"DurationUnit":"DAY"}},"WarningThreshold":50}   | arn:aws:application-signals:ap-southeast-1:<omitted>:slo/is-this-metric-source                           | 2026-05-07T06:20:11Z | CloudWatchMetric   | is-this-metric-source                           | <null>                   | {"ComparisonOperator":"","MetricThreshold":null,"RequestBasedSliMetric":{"CompositeSliConfig":null,"DependencyConfig":null,"KeyAttributes":null,"MetricSource":null,"MetricType":"","MonitoredRequestCountMetric":{"Value":[{"AccountId":null,"Expression":null,"Id":"cwMetricNumerator","Label":null,"MetricStat":{"Metric":{"Dimensions":null,"MetricName":"Errors","Namespace":"AWS/Lambda"},"Period":300,"Stat":"Average","Unit":""},"Period":null,"ReturnData":true}]},"OperationName":null,"TotalRequestCountMetric":[{"AccountId":null,"Expression":null,"Id":"cwMetricDenominator","Label":null,"MetricStat":{"Metric":{"Dimensions":null,"MetricName":"Invocations","Namespace":"AWS/Lambda"},"Period":300,"Stat":"Average","Unit":""},"Period":null,"ReturnData":true}]}}                                                                                                                                                                                                                                                                              | <null> | {}                | <null>                                                                                                      | <null>          | <null>                   | <null>          | <null>                | <null>                 | <null>                    | <null>                                                              | <null>   | <null>            | <null>   | <null>         | <null>          | <null>        | <null>                       | <null>                   | <null>                  | <null>                   | aws       | ap-southeast-1 | <omitted>  | aws                | {"connection_name":"aws","steampipe":{"sdk_version":"5.14.0"}} | {"connection_name":"aws","steampipe":{"sdk_version":"5.14.0"}} |
| arn:aws:application-signals:ap-southeast-1:<omitted>:slo/slo-test--all-operations--AVAILABILITY-97f9c797 | [{"LookBackWindowMinutes":60},{"LookBackWindowMinutes":360},{"LookBackWindowMinutes":4320}] | 2026-05-05T02:54:51Z | <null>      | RequestBased    | {"AttainmentGoal":99,"Interval":{"Value":{"Duration":28,"DurationUnit":"DAY"}},"WarningThreshold":50}    | arn:aws:application-signals:ap-southeast-1:<omitted>:slo/slo-test--all-operations--AVAILABILITY-97f9c797 | 2026-05-10T15:40:36Z | ServiceDependency  | slo-test--all-operations--AVAILABILITY-97f9c797 | <null>                   | {"ComparisonOperator":"","MetricThreshold":null,"RequestBasedSliMetric":{"CompositeSliConfig":null,"DependencyConfig":{"DependencyKeyAttributes":{"Name":"example.com","Type":"RemoteService"},"DependencyOperationName":"GET /"},"KeyAttributes":{"Environment":"lambda:default","Name":"slo-test","Type":"Service"},"MetricSource":null,"MetricType":"AVAILABILITY","MonitoredRequestCountMetric":{"Value":[{"AccountId":null,"Expression":"volume_53c54ce9510d4628bfafdb2d2c6bce86 - fault_53c54ce9510d4628bfafdb2d2c6bce86","Id":"availability_53c54ce9510d4628bfafdb2d2c6bce86","Label":null,"MetricStat":null,"Period":60,"ReturnData":true},{"AccountId":null,"Expression":null,"Id":"volume_53c54ce9510d4628bfafdb2d2c6bce86","Label":null,"MetricStat":{"Metric":{"Dimensions":[{"Name":"Service","Value":"slo-test"},{"Name":"RemoteService","Value":"example.com"},{"Name":"Environment","Value":"lambda:default"},{"Name":"RemoteOperation","Value":"GET /"}],"MetricName":"Latency","Namespace":"ApplicationSignals"},"Period":60,"Stat":"SampleCou | <null> | {"hello":"world"} | {"DependencyKeyAttributes":{"Name":"example.com","Type":"RemoteService"},"DependencyOperationName":"GET /"} | RemoteService   | <null>                   | example.com     | <null>                | <null>                 | GET /                     | {"Environment":"lambda:default","Name":"slo-test","Type":"Service"} | Service  | <null>            | slo-test | <null>         | lambda:default  | <null>        | <null>                       | <null>                   | <null>                  | <null>                   | aws       | ap-southeast-1 | <omitted>  | aws                | {"connection_name":"aws","steampipe":{"sdk_version":"5.14.0"}} | {"connection_name":"aws","steampipe":{"sdk_version":"5.14.0"}} |
|                                                                                                          |                                                                                             |                      |             |                 |                                                                                                          |                                                                                                          |                      |                    |                                                 |                          | nt","Unit":""},"Period":null,"ReturnData":false},{"AccountId":null,"Expression":null,"Id":"fault_53c54ce9510d4628bfafdb2d2c6bce86","Label":null,"MetricStat":{"Metric":{"Dimensions":[{"Name":"Service","Value":"slo-test"},{"Name":"RemoteService","Value":"example.com"},{"Name":"Environment","Value":"lambda:default"},{"Name":"RemoteOperation","Value":"GET /"}],"MetricName":"Fault","Namespace":"ApplicationSignals"},"Period":60,"Stat":"Sum","Unit":""},"Period":null,"ReturnData":false}]},"OperationName":null,"TotalRequestCountMetric":[{"AccountId":null,"Expression":null,"Id":"totalCount_53c54ce9510d4628bfafdb2d2c6bce86","Label":null,"MetricStat":{"Metric":{"Dimensions":[{"Name":"Service","Value":"slo-test"},{"Name":"RemoteService","Value":"example.com"},{"Name":"Environment","Value":"lambda:default"},{"Name":"RemoteOperation","Value":"GET /"}],"MetricName":"Latency","Namespace":"ApplicationSignals"},"Period":60,"Stat":"SampleCount","Unit":""},"Period":null,"ReturnData":true}]}}                                        |        |                   |                                                                                                             |                 |                          |                 |                       |                        |                           |                                                                     |          |                   |          |                |                 |               |                              |                          |                         |                          |           |                |            |                    |                                                                |                                                                |
| arn:aws:application-signals:ap-southeast-1:<omitted>:slo/slo-test-FunctionHandler-AVAILABILITY-9cd96247  | [{"LookBackWindowMinutes":60},{"LookBackWindowMinutes":360},{"LookBackWindowMinutes":4320}] | 2026-05-05T02:37:18Z | <null>      | RequestBased    | {"AttainmentGoal":99.99,"Interval":{"Value":{"Duration":28,"DurationUnit":"DAY"}},"WarningThreshold":50} | arn:aws:application-signals:ap-southeast-1:<omitted>:slo/slo-test-FunctionHandler-AVAILABILITY-9cd96247  | 2026-05-05T02:37:18Z | ServiceOperation   | slo-test-FunctionHandler-AVAILABILITY-9cd96247  | slo-test/FunctionHandler | {"ComparisonOperator":"","MetricThreshold":null,"RequestBasedSliMetric":{"CompositeSliConfig":null,"DependencyConfig":null,"KeyAttributes":{"Environment":"lambda:default","Name":"slo-test","Type":"Service"},"MetricSource":null,"MetricType":"AVAILABILITY","MonitoredRequestCountMetric":{"Value":[{"AccountId":null,"Expression":"volume_4192d9bf3ffb408fae6d96879520bb38 - fault_4192d9bf3ffb408fae6d96879520bb38","Id":"availability_4192d9bf3ffb408fae6d96879520bb38","Label":null,"MetricStat":null,"Period":60,"ReturnData":true},{"AccountId":null,"Expression":null,"Id":"volume_4192d9bf3ffb408fae6d96879520bb38","Label":null,"MetricStat":{"Metric":{"Dimensions":[{"Name":"Service","Value":"slo-test"},{"Name":"Environment","Value":"lambda:default"},{"Name":"Operation","Value":"slo-test/FunctionHandler"}],"MetricName":"Latency","Namespace":"ApplicationSignals"},"Period":60,"Stat":"SampleCount","Unit":""},"Period":null,"ReturnData":false},{"AccountId":null,"Expression":null,"Id":"fault_4192d9bf3ffb408fae6d96879520bb38","Label | <null> | {}                | <null>                                                                                                      | <null>          | <null>                   | <null>          | <null>                | <null>                 | <null>                    | {"Environment":"lambda:default","Name":"slo-test","Type":"Service"} | Service  | <null>            | slo-test | <null>         | lambda:default  | <null>        | <null>                       | <null>                   | <null>                  | <null>                   | aws       | ap-southeast-1 | <omitted>  | aws                | {"connection_name":"aws","steampipe":{"sdk_version":"5.14.0"}} | {"connection_name":"aws","steampipe":{"sdk_version":"5.14.0"}} |
|                                                                                                          |                                                                                             |                      |             |                 |                                                                                                          |                                                                                                          |                      |                    |                                                 |                          | ":null,"MetricStat":{"Metric":{"Dimensions":[{"Name":"Service","Value":"slo-test"},{"Name":"Environment","Value":"lambda:default"},{"Name":"Operation","Value":"slo-test/FunctionHandler"}],"MetricName":"Fault","Namespace":"ApplicationSignals"},"Period":60,"Stat":"Sum","Unit":""},"Period":null,"ReturnData":false}]},"OperationName":"slo-test/FunctionHandler","TotalRequestCountMetric":[{"AccountId":null,"Expression":null,"Id":"totalCount_4192d9bf3ffb408fae6d96879520bb38","Label":null,"MetricStat":{"Metric":{"Dimensions":[{"Name":"Service","Value":"slo-test"},{"Name":"Environment","Value":"lambda:default"},{"Name":"Operation","Value":"slo-test/FunctionHandler"}],"MetricName":"Latency","Namespace":"ApplicationSignals"},"Period":60,"Stat":"SampleCount","Unit":""},"Period":null,"ReturnData":true}]}}                                                                                                                                                                                                                               |        |                   |                                                                                                             |                 |                          |                 |                       |                        |                           |                                                                     |          |                   |          |                |                 |               |                              |                          |                         |                          |           |                |            |                    |                                                                |                                                                |
+----------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------+----------------------+-------------+-----------------+----------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------+----------------------+--------------------+-------------------------------------------------+--------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------+-------------------+-------------------------------------------------------------------------------------------------------------+-----------------+--------------------------+-----------------+-----------------------+------------------------+---------------------------+---------------------------------------------------------------------+----------+-------------------+----------+----------------+-----------------+---------------+------------------------------+--------------------------+-------------------------+--------------------------+-----------+----------------+------------+--------------------+----------------------------------------------------------------+----------------------------------------------------------------+
```
</details>

### DIFF
--- a/aws/table_aws_application_signals_service_level_objective.go
+++ b/aws/table_aws_application_signals_service_level_objective.go
@@ -1,0 +1,512 @@
+package aws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/applicationsignals"
+	"github.com/aws/aws-sdk-go-v2/service/applicationsignals/types"
+
+	"github.com/turbot/steampipe-plugin-sdk/v6/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v6/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v6/plugin/transform"
+)
+
+func tableAwsApplicationSignalsServiceLevelObjective(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_application_signals_service_level_objective",
+		Description: "AWS CloudWatch Application Signals Service Level Objective (SLO)",
+		Get: &plugin.GetConfig{
+			KeyColumns: plugin.SingleColumn("id"),
+			Hydrate:    getApplicationSignalsSlo,
+			Tags:       map[string]string{"service": "application-signals", "action": "GetServiceLevelObjective"},
+		},
+		List: &plugin.ListConfig{
+			KeyColumns: plugin.KeyColumnSlice{
+				{
+					Name:    "include_linked_accounts",
+					Require: plugin.Optional,
+				},
+				{
+					Name:    "operation_name",
+					Require: plugin.Optional,
+				},
+				{
+					Name:    "slo_owner_aws_account_id",
+					Require: plugin.Optional,
+				},
+				{
+					Name:    "metric_source_type",
+					Require: plugin.Optional,
+				},
+
+				// Fields derived from DependencyConfig
+				// Ref: https://docs.aws.amazon.com/applicationsignals/latest/APIReference/API_DependencyConfig.html
+				{Name: "dependency_type", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "dependency_resource_type", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "dependency_name", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "dependency_identifier", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "dependency_environment", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "dependency_operation_name", Require: plugin.Optional, Operators: []string{"="}},
+
+				// Fields derived from KeyAttributes
+				// Ref: https://docs.aws.amazon.com/applicationsignals/latest/APIReference/API_ListServiceLevelObjectives.html#applicationsignals-ListServiceLevelObjectives-request-KeyAttributes
+				{Name: "slo_type", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "slo_resource_type", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "slo_name", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "slo_identifier", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "slo_environment", Require: plugin.Optional, Operators: []string{"="}},
+
+				// Fields derived from MetricSource
+				// Ref: https://docs.aws.amazon.com/applicationsignals/latest/APIReference/API_MetricSource.html
+				{Name: "metric_source_key_attributes", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "metric_source_attributes", Require: plugin.Optional, Operators: []string{"="}},
+			},
+			Hydrate: listApplicationSignalsSlo,
+			Tags:    map[string]string{"service": "application-signals", "action": "ListServiceLevelObjectives"},
+		},
+		GetMatrixItemFunc: SupportedRegionMatrix(AWS_APPLICATION_SIGNALS_SERVICE_ID),
+		HydrateConfig: []plugin.HydrateConfig{
+			{
+				Func: getApplicationSignalsSloTags,
+				Tags: map[string]string{"service": "application-signals", "action": "ListTagsForResource"},
+			},
+		},
+		Columns: awsRegionalColumns([]*plugin.Column{
+			{
+				Name:        "arn",
+				Description: "The ARN of the SLO.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "burn_rate_configurations",
+				Description: "Array of configurations used to calculate the burn rate metrics of the SLO.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     getApplicationSignalsSlo,
+			},
+			{
+				Name:        "created_time",
+				Description: "The creation timestamp of the SLO.",
+				Type:        proto.ColumnType_TIMESTAMP,
+			},
+			{
+				Name:        "description",
+				Description: "The description of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Hydrate:     getApplicationSignalsSlo,
+			},
+			{
+				Name:        "evaluation_type",
+				Description: "Whether the SLO is a period-based SLO or a request-based SLO.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "goal",
+				Description: "Structure of attributes that define the goal of the SLO.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     getApplicationSignalsSlo,
+			},
+			{
+				Name:        "id",
+				Description: "The ARN of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("Arn"),
+			},
+			{
+				Name:        "last_updated_time",
+				Description: "The last update timestamp of the SLO.",
+				Type:        proto.ColumnType_TIMESTAMP,
+				Hydrate:     getApplicationSignalsSlo,
+			},
+			{
+				Name:        "metric_source_type",
+				Description: "The metric source type of the SLO.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "name",
+				Description: "The name of the SLO.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "operation_name",
+				Description: "If the SLO is specific to a single operation, this provides name of that operation.",
+				Type:        proto.ColumnType_STRING,
+			},
+			{
+				Name:        "request_based_sli",
+				Description: "If this is a request-based SLO, this contains information about the performance metric.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     getApplicationSignalsSlo,
+			},
+			{
+				Name:        "sli",
+				Description: "If this is a period-based SLO, this contains information about the performance metric.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     getApplicationSignalsSlo,
+			},
+			{
+				Name:        "tags",
+				Description: "The tags assigned to the group.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     getApplicationSignalsSloTags,
+				Transform:   transform.FromValue(),
+			},
+
+			// Attributes related to DependencyConfig
+			{
+				Name:        "dependency_config",
+				Description: "The dependency config of the SLO.",
+				Type:        proto.ColumnType_JSON,
+			},
+			{
+				Name:        "dependency_type",
+				Description: "Corresponds to DependencyConfig.DependencyKeyAttributes.Type of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DependencyConfig.DependencyKeyAttributes.Type"),
+			},
+			{
+				Name:        "dependency_resource_type",
+				Description: "Corresponds to DependencyConfig.DependencyKeyAttributes.ResourceType of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DependencyConfig.DependencyKeyAttributes.ResourceType"),
+			},
+			{
+				Name:        "dependency_name",
+				Description: "Corresponds to DependencyConfig.DependencyKeyAttributes.Name of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DependencyConfig.DependencyKeyAttributes.Name"),
+			},
+			{
+				Name:        "dependency_identifier",
+				Description: "Corresponds to DependencyConfig.DependencyKeyAttributes.Identifier of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DependencyConfig.DependencyKeyAttributes.Identifier"),
+			},
+			{
+				Name:        "dependency_environment",
+				Description: "Corresponds to DependencyConfig.DependencyKeyAttributes.Environment of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DependencyConfig.DependencyKeyAttributes.Environment"),
+			},
+			{
+				Name:        "dependency_operation_name",
+				Description: "Corresponds to DependencyConfig.DependencyOperationName of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DependencyConfig.DependencyOperationName"),
+			},
+
+			// Attributes related to KeyAttributes
+			{
+				Name:        "key_attributes",
+				Description: "A string-to-string map of key attributes of the SLO.",
+				Type:        proto.ColumnType_JSON,
+			},
+			{
+				Name:        "slo_type",
+				Description: "Corresponds to KeyAttributes.Type of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("KeyAttributes.Type"),
+			},
+			{
+				Name:        "slo_resource_type",
+				Description: "Corresponds to KeyAttributes.ResourceType of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("KeyAttributes.ResourceType"),
+			},
+			{
+				Name:        "slo_name",
+				Description: "Corresponds to KeyAttributes.Name of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("KeyAttributes.Name"),
+			},
+			{
+				Name:        "slo_identifier",
+				Description: "Corresponds to KeyAttributes.Identifier of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("KeyAttributes.Identifier"),
+			},
+			{
+				Name:        "slo_environment",
+				Description: "Corresponds to KeyAttributes.Environment of the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("KeyAttributes.Environment"),
+			},
+
+			// Attributes related to MetricSource
+			{
+				Name:        "metric_source",
+				Description: "The metric source of the SLO on resources other than Application Signals services.",
+				Type:        proto.ColumnType_JSON,
+			},
+			{
+				Name:        "metric_source_key_attributes",
+				Description: "Corresponds to MetricSource.MetricSourceKeyAttributes of the SLO.",
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("MetricSource.MetricSourceKeyAttributes"),
+			},
+			{
+				Name:        "metric_source_attributes",
+				Description: "Corresponds to MetricSource.MetricSourceAttributes of the SLO.",
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("MetricSource.MetricSourceAttributes"),
+			},
+
+			// Attributes derived from qualifiers
+			{
+				Name:        "include_linked_accounts",
+				Description: "Flag indicating whether the result set includes linked accounts.",
+				Type:        proto.ColumnType_BOOL,
+				Transform:   transform.FromQual("include_linked_accounts"),
+			},
+			{
+				Name:        "slo_owner_aws_account_id",
+				Description: "AWS account ID that owns the SLO.",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromQual("slo_owner_aws_account_id"),
+			},
+
+			// Steampipe standard columns
+			{
+				Name:        "title",
+				Description: resourceInterfaceDescription("title"),
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("Arn"),
+			},
+			{
+				Name:        "akas",
+				Description: resourceInterfaceDescription("akas"),
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("Arn").Transform(transform.EnsureStringArray),
+			},
+		}),
+	}
+}
+
+//// LIST FUNCTION
+
+func listApplicationSignalsSlo(ctx context.Context, d *plugin.QueryData, _ *plugin.HydrateData) (interface{}, error) {
+	// Get client
+	svc, err := ApplicationSignalsClient(ctx, d)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_application_signals_slo.listApplicationSignalsSlo", "client_error", err)
+		return nil, err
+	}
+
+	// Compile inputs
+	input := &applicationsignals.ListServiceLevelObjectivesInput{}
+
+	if d.EqualsQuals["include_linked_accounts"] != nil {
+		input.IncludeLinkedAccounts = d.EqualsQuals["include_linked_accounts"].GetBoolValue()
+	}
+
+	if d.EqualsQuals["operation_name"] != nil {
+		if d.EqualsQualString("operation_name") != "" {
+			input.OperationName = aws.String(d.EqualsQualString("operation_name"))
+		}
+	}
+
+	if d.EqualsQuals["slo_owner_aws_account_id"] != nil {
+		if d.EqualsQualString("slo_owner_aws_account_id") != "" {
+			input.SloOwnerAwsAccountId = aws.String(d.EqualsQualString("slo_owner_aws_account_id"))
+		}
+	}
+
+	dependencyConfig := buildDependencyConfigParam(d.Quals)
+	if dependencyConfig != nil {
+		input.DependencyConfig = dependencyConfig
+	}
+
+	keyAttributes := buildKeyAttributesParam(d.Quals)
+	if keyAttributes != nil {
+		input.KeyAttributes = *keyAttributes
+	}
+
+	paginator := applicationsignals.NewListServiceLevelObjectivesPaginator(
+		svc,
+		input,
+		func(o *applicationsignals.ListServiceLevelObjectivesPaginatorOptions) {
+			o.StopOnDuplicateToken = true
+		},
+	)
+
+	for paginator.HasMorePages() {
+		// Apply rate limiting
+		d.WaitForListRateLimit(ctx)
+
+		output, err := paginator.NextPage(ctx)
+		if err != nil {
+			plugin.Logger(ctx).Error("aws_application_signals_slo.listApplicationSignalsSlo", "api_error", err)
+			return nil, err
+		}
+
+		for _, sloSummary := range output.SloSummaries {
+			d.StreamListItem(ctx, sloSummary)
+
+			// Context may be cancelled due to manual cancellation or if the limit has been reached
+			if d.RowsRemaining(ctx) == 0 {
+				return nil, nil
+			}
+		}
+	}
+
+	return nil, nil
+}
+
+//// HYDRATE FUNCTIONS
+
+func getApplicationSignalsSlo(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+	var sloId string
+	// The SLO ID may be supplied as a parameter for a Get hydration call
+	// or as may have already been hydrated from the List hydration call.
+	if d.EqualsQualString("id") != "" {
+		sloId = d.EqualsQualString("id")
+	} else {
+		sloId = *h.Item.(types.ServiceLevelObjectiveSummary).Arn
+	}
+
+	if sloId == "" {
+		return nil, nil
+	}
+
+	// Get client
+	svc, err := ApplicationSignalsClient(ctx, d)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_application_signals_slo.getApplicationSignalsSlo", "client_error", err)
+		return nil, err
+	}
+
+	output, err := svc.GetServiceLevelObjective(ctx, &applicationsignals.GetServiceLevelObjectiveInput{
+		Id: aws.String(sloId),
+	})
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_application_signals_slo.getApplicationSignalsSlo", "api_error", err)
+		return nil, err
+	}
+
+	return output.Slo, nil
+}
+
+func getApplicationSignalsSloTags(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+	arn, err := getSloArn(ctx, d, h)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_application_signals_slo.getApplicationSignalsSloTags", "parse_error", err)
+		return nil, err
+	}
+
+	// Get client
+	svc, err := ApplicationSignalsClient(ctx, d)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_application_signals_slo.getApplicationSignalsSloTags", "client_error", err)
+		return nil, err
+	}
+
+	output, err := svc.ListTagsForResource(ctx, &applicationsignals.ListTagsForResourceInput{
+		ResourceArn: arn,
+	})
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_application_signals_slo.getApplicationSignalsSloTags", "api_error", err)
+		return nil, err
+	}
+
+	tags := make(map[string]string)
+	for _, tag := range output.Tags {
+		tags[*tag.Key] = *tag.Value
+	}
+
+	return tags, nil
+}
+
+// // UTILITY FUNCTIONS
+func buildDependencyConfigParam(quals plugin.KeyColumnQualMap) *types.DependencyConfig {
+	dependencyConfig := &types.DependencyConfig{}
+	updated := false
+
+	// Build DependencyConfig.DependencyKeyAttributes parameter
+	dependencyKeyAttributeColumnNames := map[string]string{
+		"dependency_type":          "Type",
+		"dependency_resource_type": "ResourceType",
+		"dependency_name":          "Name",
+		"dependency_identifier":    "Identifier",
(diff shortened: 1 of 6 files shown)


aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+		if quals[qualName] == nil {
+			continue
+		}
+		for _, q := range quals[qualName].Quals {
+			value := q.Value.GetStringValue()
+			if value == "" || q.Operator != "=" {
+				continue
+			}
+
+			dependencyConfig.DependencyKeyAttributes[paramName] = value
```
cbruno10, line 587: Does this function work as intended? It seems like because  `dependencyConfig` is built with `&types.DependencyConfig{}` at the start of the function, this leaves `DependencyKeyAttributes` as a nil map, so any `dependency_*` qual hits `assignment to entry in nil map`. Do we need to initialize it first, e.g.,
```go
dependencyConfig := &types.DependencyConfig{
    DependencyKeyAttributes: make(map[string]string),
}
```

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+	"github.com/turbot/steampipe-plugin-sdk/v6/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v6/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v6/plugin/transform"
+)
+
+func tableAwsApplicationSignalsServiceLevelObjective(_ context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "aws_application_signals_service_level_objective",
+		Description: "AWS CloudWatch Application Signals Service Level Objective (SLO)",
+		Get: &plugin.GetConfig{
```
cbruno10, line 20: I think we need an `IgnoreConfig`, similar to https://github.com/turbot/steampipe-plugin-aws/blob/5b4b0511f4251a99294e1fe5134f73068ae78694/aws/table_aws_ec2_application_load_balancer.go#L22-L24

For instance:
```go
IgnoreConfig: &plugin.IgnoreConfig{
    ShouldIgnoreErrorFunc: shouldIgnoreErrors([]string{"ResourceNotFoundException", "ValidationException"}),
},
```

I'm not sure if the Get API call can return `ValidationException`, but worth verifying.

I think we also need the `IgnoreConfig` in the tags hydrate in `HydrateConfig`.

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+				// Ref: https://docs.aws.amazon.com/applicationsignals/latest/APIReference/API_ListServiceLevelObjectives.html#applicationsignals-ListServiceLevelObjectives-request-KeyAttributes
+				{Name: "slo_type", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "slo_resource_type", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "slo_name", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "slo_identifier", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "slo_environment", Require: plugin.Optional, Operators: []string{"="}},
+
+				// Fields derived from MetricSource
+				// Ref: https://docs.aws.amazon.com/applicationsignals/latest/APIReference/API_MetricSource.html
+				{Name: "metric_source_key_attributes", Require: plugin.Optional, Operators: []string{"="}},
```
cbruno10: These three key columns (`metric_source_type`, `metric_source_key_attributes` and `metric_source_attributes` here) are set as quals but never used in the request in  `listApplicationSignalsSlo`. Are these meant to be wired in?

Also, in general, we don't need `Operators: []string{"="}` if the only operator is `=`, as that's the default.

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromQual("slo_owner_aws_account_id"),
+			},
+
+			// Steampipe standard columns
+			{
+				Name:        "title",
+				Description: resourceInterfaceDescription("title"),
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("Arn"),
```
cbruno10: Can we use the SLO's name instead of ARN, as it's more friendly?

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+func listApplicationSignalsSlo(ctx context.Context, d *plugin.QueryData, _ *plugin.HydrateData) (interface{}, error) {
+	// Get client
+	svc, err := ApplicationSignalsClient(ctx, d)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_application_signals_slo.listApplicationSignalsSlo", "client_error", err)
+		return nil, err
+	}
+
+	// Compile inputs
+	input := &applicationsignals.ListServiceLevelObjectivesInput{}
```
cbruno10, line 390: Can you please add a clamping block here as `ListServiceLevelObjectivesInput` supports `MaxResults` (default 50), e.g.,
```go
maxLimit := int32(50)
if d.QueryContext.Limit != nil {
    limit := int32(*d.QueryContext.Limit)
    if limit < maxLimit {
        if limit < 1 {
            maxLimit = 1
        } else {
            maxLimit = limit
        }
    }
}
input.MaxResults = aws.Int32(maxLimit)
```

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+	}
+}
+
+//// LIST FUNCTION
+
+func listApplicationSignalsSlo(ctx context.Context, d *plugin.QueryData, _ *plugin.HydrateData) (interface{}, error) {
+	// Get client
+	svc, err := ApplicationSignalsClient(ctx, d)
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_application_signals_slo.listApplicationSignalsSlo", "client_error", err)
```
cbruno10:
```suggestion
		plugin.Logger(ctx).Error("aws_application_signals_service_level_objective.listApplicationSignalsSlo", "client_error", err)
```

For logs, we usually use `table_name.function` for the first part.

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+
+	tags := make(map[string]string)
+	for _, tag := range output.Tags {
+		tags[*tag.Key] = *tag.Value
+	}
+
+	return tags, nil
+}
+
+// // UTILITY FUNCTIONS
```
cbruno10:
```suggestion
//// UTILITY FUNCTIONS
```

docs/tables/aws_application_signals_service_level_objective.md
```diff
@@ -0,0 +1,44 @@
...
+### Fetch Service Level Objectives Based on CloudWatch Metrics
+Fetch service level objectives configured based on CloudWatch metrics.
+
+```sql+postgres
+select
+  name,
+  burn_rate_configurations,
+  goal
+from
+  aws_application_signals_slo
```
cbruno10: This table name doesn't exist. The registered table is `aws_application_signals_service_level_objective`. Same fix needed in the `sql+sqlite` block at L41. Also, both queries should end with a semicolon.

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+			},
+			{
+				Name:        "sli",
+				Description: "If this is a period-based SLO, this contains information about the performance metric.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     getApplicationSignalsSlo,
+			},
+			{
+				Name:        "tags",
+				Description: "The tags assigned to the group.",
```
cbruno10:
```suggestion
				Description: resourceInterfaceDescription("tags"),
```

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+				Hydrate:     getApplicationSignalsSlo,
+			},
+			{
+				Name:        "sli",
+				Description: "If this is a period-based SLO, this contains information about the performance metric.",
+				Type:        proto.ColumnType_JSON,
+				Hydrate:     getApplicationSignalsSlo,
+			},
+			{
+				Name:        "tags",
```
cbruno10: Can we move this `tags` column below the `title` column under `// Steampipe standard columns`?

cbruno10: Can you please add a `tags_src` column, similar to those in EC2 tables, like https://github.com/turbot/steampipe-plugin-aws/blob/5b4b0511f4251a99294e1fe5134f73068ae78694/aws/table_aws_ec2_application_load_balancer.go#L140

We add these when we transform the tags we get from the API

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+				// Fields derived from MetricSource
+				// Ref: https://docs.aws.amazon.com/applicationsignals/latest/APIReference/API_MetricSource.html
+				{Name: "metric_source_key_attributes", Require: plugin.Optional, Operators: []string{"="}},
+				{Name: "metric_source_attributes", Require: plugin.Optional, Operators: []string{"="}},
+			},
+			Hydrate: listApplicationSignalsSlo,
+			Tags:    map[string]string{"service": "application-signals", "action": "ListServiceLevelObjectives"},
+		},
+		GetMatrixItemFunc: SupportedRegionMatrix(AWS_APPLICATION_SIGNALS_SERVICE_ID),
+		HydrateConfig: []plugin.HydrateConfig{
```
cbruno10, line 73: Can you please also add the get function here in the `HydrateConfig`, similar to https://github.com/turbot/steampipe-plugin-aws/blob/5b4b0511f4251a99294e1fe5134f73068ae78694/aws/table_aws_codeartifact_domain.go#L49-L52, to allow throttling to work correctly?

docs/tables/aws_application_signals_service_level_objective.md
```diff
@@ -0,0 +1,44 @@
...
+## Table Usage Guide
+
+The `aws_application_signals_service_level_objective` table in Steampipe provides you with information about CloudWatch
+Application Signals service level objectives. This table allows you, as a DevOps engineer, to configure service level
+objectives based on a variety of metrics and monitor your applications' performance. You can utilize this table to
+manage and monitor service level objectives of your applications. The schema outlines the various attributes of the
+service level objectives for you, including their metric sources and alarm thresholds.
+
+## Examples
+
```
cbruno10, line 20: Can you please add a `### Basic info` example here, like most other tables have (https://steampipe.io/docs/develop/table-docs-standards#example-query-guidelines)?

docs/tables/aws_application_signals_service_level_objective.md
```diff
@@ -0,0 +1,44 @@
...
+### Fetch Service Level Objectives Based on CloudWatch Metrics
```
cbruno10:
```suggestion
### List service level objectives based on CloudWatch metrics
```

Can you please update all section titles to use sentence case and start with an imperative? More details in https://steampipe.io/docs/develop/table-docs-standards#style-conventions-1

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,512 @@
...
+			{
+				Name:        "metric_source",
+				Description: "The metric source of the SLO on resources other than Application Signals services.",
+				Type:        proto.ColumnType_JSON,
+			},
+			{
+				Name:        "metric_source_key_attributes",
+				Description: "Corresponds to MetricSource.MetricSourceKeyAttributes of the SLO.",
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("MetricSource.MetricSourceKeyAttributes"),
```
cbruno10: `MetricSource` lives at a different path in the List vs. Get responses. It's a top-level field on `ServiceLevelObjectiveSummary` (the list item), but on `ServiceLevelObjective` (the get item) it's nested under `Sli.SliMetric` / `RequestBasedSli.RequestBasedSliMetric`. So with the current single-path transform this column populates on a full scan but comes back null on a `where id = '...'` lookup.

I think we can use `transform.FromField` with multiple paths to help resolve the data:

```suggestion
				Transform: transform.FromField("MetricSource.MetricSourceKeyAttributes",
					"Sli.SliMetric.MetricSource.MetricSourceKeyAttributes",
					"RequestBasedSli.RequestBasedSliMetric.MetricSource.MetricSourceKeyAttributes"),
```

The same fix applies to the other columns sourced from `KeyAttributes`, `DependencyConfig`, `MetricSource`, and `OperationName`. I had a quick look at the API response data and believe this is correct, but worth testing live by checking each column data with a list vs. get call.

cbruno10 (changes requested):
@leongzhiyong Thanks for another table PR! Can you please have a look at the suggestions and let us know if you have any questions? Thanks!

aws/table_aws_application_signals_service_level_objective.go
leongzhiyong: Thanks for catching that, I missed out on test coverage for the dependency config.

aws/table_aws_application_signals_service_level_objective.go
leongzhiyong: It can, according to the [docs](https://docs.aws.amazon.com/applicationsignals/latest/APIReference/API_GetServiceLevelObjective.html) when the ARN format is invalid.

`ListTagsForResource` only throws `ResourceNotFoundException`.

aws/table_aws_application_signals_service_level_objective.go
leongzhiyong: Thanks, missed those when I renamed the table to use the long-form.

aws/table_aws_application_signals_service_level_objective.go
leongzhiyong: Adjusted the hydrate functions for `tags` and `tags_src` to return the Turbot-standard map and original response values respectively,  as `table_aws_ec2_application_load_balancer.go` does.

aws/table_aws_application_signals_service_level_objective.go
leongzhiyong: Have done so. This is interesting to me.

Is the premise that because `getApplicationSignalsSlo` is used not just to hydrate the Get operation but also additional attributes (e.g. `burn_rate_configurations` , `description`), we also need to register it with `HydrateConfig`?

cbruno10: @leongzhiyong - Yes, that's the premise. When the get function is also used as a column hydrate (`description`, `burn_rate_configurations`, etc.), those calls run per row during list queries, and the service/action tags are what the rate limiter uses to throttle them.

aws/table_aws_application_signals_service_level_objective.go
leongzhiyong: Oops, yep I did mean to wire them in.

aws/table_aws_application_signals_service_level_objective.go
leongzhiyong: Thanks for catching that. I've checked all the response attributes to find those that this behaviour applies to, made the adjustments to their respective transforms, and tested both the Sli and RequestBasedSli types with both List and Get calls.

Disclaimer: I couldn't quite figure out how to create SLOs that contain certain attributes e.g. `Sli.SliMetric.MetricSource.MetricSourceAttributes` so the implementation for those are...strictly theoretical.

leongzhiyong:
All good for another round!

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,676 @@
...
+
+	if d.EqualsQualString("slo_owner_aws_account_id") != "" {
+		input.SloOwnerAwsAccountId = aws.String(d.EqualsQualString("slo_owner_aws_account_id"))
+	}
+
+	if d.EqualsQualString("metric_source_type") != "" {
+		input.MetricSourceTypes = []types.MetricSourceType{types.MetricSourceType(d.EqualsQualString("metric_source_type"))}
+	}
+
+	dependencyConfig := buildDependencyConfigParam(d.Quals)
```
cbruno10: The dependency filters can't be used on their own. I ran the branch against an account with an SLO and got:

- `where dependency_type = 'RemoteService'` fails in the SDK before any request: `missing required field, DependencyConfig.DependencyOperationName`
- `where dependency_operation_name = 'GET /'` alone gets a 400: `dependencyKeyAttributes ... Member must have length greater than or equal to 1`
- both together still 400: `KeyAttributes must be provided when DependencyConfig is provided`

So a dependency filter needs the operation name, at least one `dependency_*` attribute, and the `slo_*` key attributes all at once. Could we only set `DependencyConfig` when the operation name and at least one attribute are present, and spell out the full combination in the docs? Otherwise a single `dependency_type` qual is a guaranteed error.

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,676 @@
...
+	}
+
+	metricSourceKeyAttributesParam, err := buildParamsFromJson(d.Quals, "metric_source_key_attributes")
+	if err != nil {
+		plugin.Logger(ctx).Error("aws_application_signals_service_level_objective.listApplicationSignalsSlo", "parse_error", err)
+		return nil, err
+	}
+
+	if len(metricSourceKeyAttributesParam) > 0 || len(metricSourceAttributesParam) > 0 {
+		input.MetricSource = &types.MetricSource{}
```
cbruno10: Same shape here. `where metric_source_attributes = '{...}'` on its own fails client side with `missing required field, MetricSource.MetricSourceKeyAttributes`. Suggest only building `MetricSource` when `metric_source_key_attributes` is present, and treating `metric_source_attributes` as an add-on to it.

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,676 @@
...
+	if d.EqualsQuals["include_linked_accounts"] != nil {
+		input.IncludeLinkedAccounts = d.EqualsQuals["include_linked_accounts"].GetBoolValue()
+	}
+
+	if d.EqualsQualString("operation_name") != "" {
+		input.OperationName = aws.String(d.EqualsQualString("operation_name"))
+	}
+
+	if d.EqualsQualString("slo_owner_aws_account_id") != "" {
+		input.SloOwnerAwsAccountId = aws.String(d.EqualsQualString("slo_owner_aws_account_id"))
```
cbruno10, line 414: A query with `where slo_owner_aws_account_id = '...'` on its own returns a 400 error: `IncludeLinkedAccounts must be 'true' when specifying SloOwnerAwsAccountId`. Since the one implies the other, can we set `input.IncludeLinkedAccounts = true` whenever the `slo_owner_aws_account_id` qual is given?

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,676 @@
...
+				Type:        proto.ColumnType_STRING,
+				Transform: transform.FromField(
+					"KeyAttributes.Environment",
+					"Sli.SliMetric.KeyAttributes.Environment",
+					"RequestBasedSli.RequestBasedSliMetric.KeyAttributes.Environment",
+				),
+			},
+			{
+				Name:        "tags_src",
+				Description: "A list of tags assigned to the group.",
```
cbruno10, line 303:
```suggestion
				Description: "A list of tags assigned to the service level objective.",
```

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,676 @@
...
+func getSloArn(_ context.Context, _ *plugin.QueryData, h *plugin.HydrateData) (*string, error) {
+	if h.Item != nil {
+		switch item := h.Item.(type) {
+		case *types.ServiceLevelObjective:
+			return item.Arn, nil
+		case types.ServiceLevelObjectiveSummary:
+			return item.Arn, nil
+		}
+	}
+	return nil, nil
```
cbruno10, line 675: If we ever get here we go on to call `ListTagsForResource` with a nil ARN. Both current item types are covered, but should we return an error instead of `nil, nil`?

go.mod
```diff
@@ -168,6 +168,7 @@ require (
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.55.0 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/internal/resourcemapping v0.55.0 // indirect
 	github.com/aws/aws-sdk-go-v2/credentials v1.19.12 // indirect
+	github.com/aws/aws-sdk-go-v2/service/applicationsignals v1.22.4 // indirect
```
cbruno10, line 171: Can you please run `go mod tidy`, which should clean up some indirect dependencies I think?

docs/tables/aws_application_signals_service_level_objective.md
```diff
@@ -0,0 +1,73 @@
...
+  evaluation_type,
+  metric_source_type,
+  created_time,
+  last_updated_time,
+  description
+from
+  aws_application_signals_service_level_objective;
+```
+
+### List service level objectives based on CloudWatch metrics
```
cbruno10, line 50: Ran this against a real SLO and list, get by ARN, `tags`, `tags_src`, `akas` and the not-found path all behaved, nice.

Once the attribute quals are out, could you add an example for each of the remaining ones (`metric_source_type`, `operation_name`, `include_linked_accounts`, and `slo_owner_aws_account_id` together with `include_linked_accounts = true`), and post the query output for each in the PR description so we can see them working as intended? The current output there predates `tags_src`, `title` and `akas` as well.

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,676 @@
...
+					Require: plugin.Optional,
+				},
+				{
+					Name:    "operation_name",
+					Require: plugin.Optional,
+				},
+				{
+					Name:    "slo_owner_aws_account_id",
+					Require: plugin.Optional,
+				},
```
cbruno10, line 41:
```suggestion
				{
					Name:       "include_linked_accounts",
					Require:    plugin.Optional,
					CacheMatch: query_cache.CacheMatchExact,
				},
				{
					Name:    "operation_name",
					Require: plugin.Optional,
				},
				{
					Name:       "slo_owner_aws_account_id",
					Require:    plugin.Optional,
					CacheMatch: query_cache.CacheMatchExact,
				},
```

These two are request parameters rather than properties of the SLO, and their columns come from `FromQual`, so they need exact cache matching. With the default subset match, an unqualified query gets cached with these columns null, and a later `where include_linked_accounts = true` is served from that cache and filtered against null, so it returns nothing instead of the linked-account SLOs. This change will also need the `query_cache` import as well.

aws/table_aws_application_signals_service_level_objective.go
```diff
@@ -0,0 +1,676 @@
...
+				{Name: "slo_type", Require: plugin.Optional},
+				{Name: "slo_resource_type", Require: plugin.Optional},
+				{Name: "slo_name", Require: plugin.Optional},
+				{Name: "slo_identifier", Require: plugin.Optional},
+				{Name: "slo_environment", Require: plugin.Optional},
+
+				// Fields derived from MetricSource
+				// Ref: https://docs.aws.amazon.com/applicationsignals/latest/APIReference/API_MetricSource.html
+				{Name: "metric_source_key_attributes", Require: plugin.Optional},
+				{Name: "metric_source_attributes", Require: plugin.Optional},
```
cbruno10, line 67:
```suggestion
```

Having tried these, I think we should drop the attribute quals altogether and keep the columns as plain columns. The API only accepts them in specific combinations: a `dependency_*` filter needs the operation name, at least one dependency attribute and the `slo_*` key attributes all at once, `metric_source_attributes` needs `metric_source_key_attributes`, and `slo_type = 'Service'` on its own is rejected because Service also needs Name. On the branch a single one of these quals is almost always a 400, and the pairing logic is where the last two bugs came from.

SLOs are a small resource per account, so listing everything and filtering in SQL costs very little, and all of this data comes back on the list response anyway. That would also let `buildDependencyConfigParam`, `buildKeyAttributesParam`, `buildParamsFromJson` and the matching blocks in `listApplicationSignalsSlo` go, and we wouldn't need a docs section explaining the combinations.

Sorry for the back and forth on this one, the earlier ask to wire these up was mine. Keeping `include_linked_accounts`, `slo_owner_aws_account_id`, `metric_source_type` and `operation_name` as quals still makes sense since those are real request parameters.

cbruno10 (changes requested):
@leongzhiyong Sorry for the long response time! I tried this table again and ran into a few issues trying to use some of the quals, I think largely removing a lot of them should still keep the table fast, as there often aren't a lot of these resource types, but open to discussion on that. Thanks!

PR #2772, opened 2026-06-13
feat: Add `presigned_url` column and `start_after` pagination to `aws_s3_object`

OsMikeBl:
## Summary

Adds two enhancements to the `aws_s3_object` table:

- **`presigned_url` column** — generates a temporary pre-authenticated download URL for each object. Expiration is controlled via `presigned_url_expires_in` (default 900s / 15 minutes).
- **`start_after` column** — enables cursor-based pagination over large buckets. Maps directly to the S3 `ListObjectsV2` `StartAfter` parameter. Combined with `LIMIT`, allows efficient page-by-page traversal without loading the entire bucket.

## Usage examples

```sql
-- Generate presigned URLs (valid 1 hour)
select key, presigned_url
from aws_s3_object
where bucket_name = 'my-bucket'
  and prefix = 'reports/'
  and presigned_url_expires_in = 3600;

-- Paginate through a large bucket
select key from aws_s3_object
where bucket_name = 'huge-bucket' limit 1000;

-- Next page: use last key from previous result
select key from aws_s3_object
where bucket_name = 'huge-bucket'
  and start_after = 'last-key-from-previous-page'
limit 1000;
```

## Test plan

- [ ] `go build ./...` — compiles cleanly
- [ ] Query with `presigned_url` column returns valid, working URLs
- [ ] Query with `presigned_url_expires_in = 3600` generates URLs with correct expiry
- [ ] Query with `start_after` returns objects lexicographically after the specified key
- [ ] Query without new columns behaves identically to before (no regression)
- [ ] Presigned URL errors (e.g., insufficient permissions) return clean error, not panic

### DIFF
--- a/aws/table_aws_s3_object.go
+++ b/aws/table_aws_s3_object.go
@@ -5,7 +5,6 @@ import (
 	"encoding/base64"
 	"io"
 	"strings"
-	"time"
 	"unicode/utf8"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -27,7 +26,7 @@ func tableAwsS3Object(_ context.Context) *plugin.Table {
 			KeyColumns: []*plugin.KeyColumn{
 				{Name: "bucket_name", Require: plugin.Required, CacheMatch: query_cache.CacheMatchExact},
 				{Name: "prefix", Require: plugin.Optional, CacheMatch: query_cache.CacheMatchExact},
-				{Name: "start_after", Require: plugin.Optional},
+				{Name: "start_after", Require: plugin.Optional, CacheMatch: query_cache.CacheMatchExact},
 
 				// If you encrypt an object by using server-side encryption with customer-provided
 				// encryption keys (SSE-C) when you store the object in Amazon S3, then when you
@@ -35,9 +34,6 @@ func tableAwsS3Object(_ context.Context) *plugin.Table {
 				{Name: "sse_customer_algorithm", Require: plugin.Optional},
 				{Name: "sse_customer_key", Require: plugin.Optional, CacheMatch: query_cache.CacheMatchExact},
 				{Name: "sse_customer_key_md5", Require: plugin.Optional},
-
-				// Presigned URL expiration in seconds (default: 900)
-				{Name: "presigned_url_expires_in", Require: plugin.Optional},
 			},
 		},
 
@@ -71,10 +67,6 @@ func tableAwsS3Object(_ context.Context) *plugin.Table {
 				Depends: []plugin.HydrateFunc{getBucketRegionForObjects},
 				Tags:    map[string]string{"service": "s3", "action": "GetObjectTagging"},
 			},
-			{
-				Func: getS3ObjectPresignedURL,
-				Tags: map[string]string{"service": "s3", "action": "PresignGetObject"},
-			},
 		},
 		Columns: awsAccountColumns([]*plugin.Column{
 			{
@@ -97,7 +89,7 @@ func tableAwsS3Object(_ context.Context) *plugin.Table {
 			},
 			{
 				Name:        "start_after",
-				Description: "When set, listing starts after this key in lexicographic order. Use the last key from a previous LIMIT query to paginate. Only works with default key ordering (no ORDER BY).",
+				Description: "Amazon S3 starts listing after this specified key. StartAfter can be any key in the bucket.",
 				Type:        proto.ColumnType_STRING,
 				Transform:   transform.FromQual("start_after"),
 			},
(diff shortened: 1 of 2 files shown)


aws/table_aws_s3_object.go
```diff
@@ -86,6 +95,12 @@ func tableAwsS3Object(_ context.Context) *plugin.Table {
 				Type:        proto.ColumnType_STRING,
 				Transform:   transform.FromQual("bucket_name"),
 			},
+			{
+				Name:        "start_after",
+				Description: "When set, listing starts after this key in lexicographic order. Use the last key from a previous LIMIT query to paginate. Only works with default key ordering (no ORDER BY).",
```
cbruno10:
```suggestion
				Description: "Amazon S3 starts listing after this specified key. StartAfter can be any key in the bucket.",
```

For guidance on how to use the column or quals, we usually put those in the docs if there's anything specific to note (not sure there is in this case, as the current description describes general SQL behaviour).

aws/table_aws_s3_object.go
```diff
@@ -26,13 +27,17 @@ func tableAwsS3Object(_ context.Context) *plugin.Table {
 			KeyColumns: []*plugin.KeyColumn{
 				{Name: "bucket_name", Require: plugin.Required, CacheMatch: query_cache.CacheMatchExact},
 				{Name: "prefix", Require: plugin.Optional, CacheMatch: query_cache.CacheMatchExact},
+				{Name: "start_after", Require: plugin.Optional},
```
cbruno10:
```suggestion
				{Name: "start_after", Require: plugin.Optional, CacheMatch: query_cache.CacheMatchExact},
```

Without exact cache match, with caching enabled, if you run an unqualified  query, and then another right after with `start_after` set, you won't get the correct results.

aws/table_aws_s3_object.go
```diff
@@ -330,6 +345,19 @@ func tableAwsS3Object(_ context.Context) *plugin.Table {
 				Transform:   transform.FromField("WebsiteRedirectLocation"),
 				Hydrate:     headS3Object,
 			},
+			{
+				Name:        "presigned_url",
```
cbruno10: Steampipe tables report on the state of resources, and a presigned URL isn't a property of the object, it's a short-lived credential you generate to share or download a file. AWS treats it that way too (it's a separate `aws s3 presign` / `PresignGetObject` operation, not part of any list call), and generating one per row would bake temporary credentials into the query cache and any exports.

So I think for presigned URLs, we would most likely not add this capability into the `aws_s3_object` table, and as a workaround, use the AWS CLI, e.g., `aws s3 presign s3://my-bucket/reports/2024-q1.csv --expires-in 3600`, as a follow-up to any Steampipe queries.

aws/table_aws_s3_object.go
```diff
@@ -429,6 +457,8 @@ func listS3Objects(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateDa
 		return nil, err
 	}
 
+	equalQuals := d.EqualsQuals
```
cbruno10: Any reason this moved? Functionally it's the same, but declaring it near where it's used seems helpful

cbruno10 (changes requested):
@OsMikeBl Thanks for raising the PR! I've left a few review comments, can you please take a look when you get a chance?

aws/table_aws_s3_object.go
OsMikeBl: @cbruno10 , fixed in [1221c75](https://github.com/turbot/steampipe-plugin-aws/pull/2772/commits/1221c755d0d43fb743f66bb12ef3dd9cd1989a1e) 🙏

aws/table_aws_s3_object.go
OsMikeBl: @cbruno10 , fixed in [1221c75](https://github.com/turbot/steampipe-plugin-aws/pull/2772/commits/1221c755d0d43fb743f66bb12ef3dd9cd1989a1e) 🙏

aws/table_aws_s3_object.go
OsMikeBl: @cbruno10 , that makes sense. My motivation was a workflow where Steampipe has AWS credentials but the caller consuming the query results does not, so a presigned URL would let the caller download or stream large objects without fetching body through Steampipe. Removed in [1221c75](https://github.com/turbot/steampipe-plugin-aws/pull/2772/commits/1221c755d0d43fb743f66bb12ef3dd9cd1989a1e)

aws/table_aws_s3_object.go
OsMikeBl: @cbruno10 , fixed in [1221c75](https://github.com/turbot/steampipe-plugin-aws/pull/2772/commits/1221c755d0d43fb743f66bb12ef3dd9cd1989a1e) 🙏

cbruno10 (approved):

cbruno10:
Thanks @OsMikeBl for making the changes! I've merged the PR in and we'll include this change in our next release, which will be next week.

PR #2786, opened 2026-08-26
Add tag_filter key column support to aws_tagging_resource table

asreenath:
## Description
Adds `tag_filter` key column support to the `aws_tagging_resource` table, enabling users to filter AWS resources by tags directly in Steampipe queries.

## Changes
- Added `tag_filter` JSON key column (optional, cache exact match)
- Implemented `buildTagFilter()` helper to parse JSON and construct AWS SDK TagFilter array
- Added 11 unit tests covering all scenarios (valid inputs, invalid JSON, edge cases)
- Added integration test files for end-to-end validation

## Example Usage
```sql
-- Filter by single tag key (matches any value)
SELECT * FROM aws_tagging_resource 
WHERE tag_filter = '[{"key":"Environment"}]';

-- Filter by tag key with specific values
SELECT * FROM aws_tagging_resource 
WHERE tag_filter = '[{"key":"Environment","values":["prod","staging"]}]';

-- Filter by multiple tags (AND logic)
SELECT * FROM aws_tagging_resource 
WHERE tag_filter = '[{"key":"Team","values":["platform"]},{"key":"Environment"}]';
```

## Testing
- ✅ Unit tests pass: `go test ./aws -run TestBuildTagFilter`
- ✅ Build succeeds: `go build`
- ✅ Full test suite passes: `go test ./aws`
- ✅ Integration tests pass: All 5 test cases passed

## Implementation Details
- JSON format matches AWS SDK v2 `types.TagFilter` structure
- Pattern follows existing `inspector2_finding` table's `buildResourceTagFilter` approach
- Error handling returns clear messages for invalid JSON input
- Uses SDK v6 `quals.QualSlice` for type-safe qualifier handling

# Integration test logs
<details>
  <summary>Logs</summary>

```
SETUP: tests/aws_tagging_resource []

PRETEST: tests/aws_tagging_resource

TEST: tests/aws_tagging_resource
Running terraform
data.aws_region.alternate: Reading...
data.aws_region.alternate: Read complete after 0s [id=us-east-2]
data.aws_region.primary: Reading...
data.aws_caller_identity.current: Reading...
data.aws_partition.current: Reading...
data.aws_partition.current: Read complete after 0s [id=aws]
data.aws_region.primary: Read complete after 0s [id=us-east-1]
data.aws_caller_identity.current: Read complete after 0s [id=<MYACCT>]
data.null_data_source.resource: Reading...
data.null_data_source.resource: Read complete after 0s [id=static]

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create
 <= read (data resources)

Terraform will perform the following actions:

  # data.aws_resourcegroupstaggingapi_resources.named_test_resource will be read during apply
  # (depends on a resource or a module with changes pending)
 <= data "aws_resourcegroupstaggingapi_resources" "named_test_resource" {
      + id                        = (known after apply)
      + region                    = (known after apply)
      + resource_tag_mapping_list = (known after apply)

      + tag_filter {
          + key    = "Name"
          + values = [
              + "turbottest14435",
            ]
        }
    }

  # aws_ebs_volume.named_test_resource will be created
  + resource "aws_ebs_volume" "named_test_resource" {
      + arn               = (known after apply)
      + availability_zone = "us-east-1a"
      + encrypted         = (known after apply)
      + final_snapshot    = false
      + id                = (known after apply)
      + iops              = (known after apply)
      + size              = 1
      + tags              = {
          + "Name" = "turbottest14435"
        }
      + tags_all          = {
          + "Name" = "turbottest14435"
        }
      + type              = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

aws_ebs_volume.named_test_resource: Creating...
aws_ebs_volume.named_test_resource: Still creating... [00m10s elapsed]
aws_ebs_volume.named_test_resource: Creation complete after 11s [id=vol-09653c2f252238b78]
data.aws_resourcegroupstaggingapi_resources.named_test_resource: Reading...
data.aws_resourcegroupstaggingapi_resources.named_test_resource: Read complete after 0s [id=aws]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.

Outputs:

account_id = "<MYACCT>"
aws_partition = "aws"
aws_region = "us-east-1"
resource_aka = "arn:aws:ec2:us-east-1:<MYACCT>:volume/vol-09653c2f252238b78"
resource_id = "vol-09653c2f252238b78"
resource_name = "turbottest14435"
resource_tags = tomap({
  "Name" = "turbottest14435"
})
resource_title = "volume/vol-09653c2f252238b78"

Running SQL query: test-get-query.sql
✔ PASSED

Running SQL query: test-list-query.sql
✔ PASSED

Running SQL query: test-notfound-query.sql
✔ PASSED

Running SQL query: test-tag-filter-query.sql
✔ PASSED

Running SQL query: test-turbot-query.sql
✔ PASSED

POSTTEST: tests/aws_tagging_resource

TEARDOWN: tests/aws_tagging_resource

SUMMARY:

1/1 passed.
```
</details>

# Example query results
<details>
  <summary>Results</summary>

```sql
-- Test Query: Filter by tag key with specific value
SELECT name, arn, tags
FROM aws_tagging_resource
WHERE tag_filter = '[{"key":"Name","values":["turbottest14435"]}]'::jsonb;
```

**Result:**
```json
[
  {
    "arn": "arn:aws:ec2:us-east-1:<MYACCT>:volume/vol-09653c2f252238b78",
    "name": "volume/vol-09653c2f252238b78",
    "tags": {
      "Name": "turbottest14435"
    }
  }
]
```

**Additional test queries executed:**

```sql
-- Get resource by ARN
SELECT account_id, akas, arn, name, partition, region, tags
FROM aws_tagging_resource
WHERE resource_arn = 'arn:aws:ec2:us-east-1:<MYACCT>:volume/vol-09653c2f252238b78';

-- List resources
SELECT arn, name
FROM aws_tagging_resource;

-- Handle not-found case
SELECT *
FROM aws_tagging_resource
WHERE resource_arn = 'arn:aws:ec2:us-east-1:<MYACCT>:volume/vol-nonexistent';
```

All queries returned expected results. The tag_filter functionality correctly filters resources by tags using the JSON format matching AWS SDK v2 TagFilter structure.
</details>

## Checklist
- [x] Code follows repository patterns
- [x] Unit tests added and passing
- [x] Integration test files added
- [x] Integration tests executed and passing
- [x] Documentation/examples included in code comments
- [x] Tested with Go 1.27.0
- [x] Infrastructure cleanup verified


asreenath:
## Response to Review Comments

### Comment 1: Multiple Qualifiers Concern

Thank you for the detailed review. However, I believe this concern doesn't apply to this implementation:

**Why multiple qualifiers won't occur:**
1. The `tag_filter` column only supports the `=` operator (not `IN` or other operators that could create multiple qualifiers)
2. The column type is JSONB, which accepts a single JSON value
3. Users cannot write queries that would create multiple qualifiers for this column

**Current behavior is correct:**
When users provide a JSON array like `[{"key":"Env","values":["prod"]},{"key":"Team"}]`, they are specifying a single filter that means "resources with BOTH Env=prod AND Team tags". The AWS API correctly ANDs these TagFilter objects together, which matches the user's intent.

**OR semantics are already supported:**
Users can express OR logic within a single TagFilter using the values array:
```sql
WHERE tag_filter = '[{"key":"Env","values":["prod","staging"]}]'
```
This returns resources with Env=prod OR Env=staging, as expected.

### Comment 2: Formatting

Fixed in commit 357fb8f0:
- ✅ Updated JSON example to use double quotes (valid JSON)
- ✅ Ran gofmt to fix indentation

asreenath:
## Response to Comment 3: Composite Literal Syntax

The code **does compile** and all tests pass. The syntax is valid Go.

**Evidence:**
- ✅ Tests compile: `go test -c ./aws -run TestBuildTagFilter` produces no errors
- ✅ All 11 tests pass: `go test ./aws -run TestBuildTagFilter -v`

**Both forms are valid Go:**
```go
// Shorthand (type inferred) - VALID
map[string][]string{"key": {"val1", "val2"}}

// Explicit (type stated) - ALSO VALID
map[string][]string{"key": []string{"val1", "val2"}}
```

**Official Go documentation confirms shorthand is acceptable:**
- [Effective Go](https://go.dev/doc/effective_go) shows shorthand examples
- [Go Language Spec](https://go.dev/ref/spec#Composite_literals) explicitly permits type elision when type can be inferred from context
- [Google Go Style Guide](https://google.github.io/styleguide/go/best-practices.html) prefers shorthand for readability
- Go standard library uses shorthand extensively (e.g., `net/http/client_test.go`: `Header: map[string][]string{"Location": {"http://..."}}`)

**No Go linter flags this as an issue** (staticcheck, go vet).

The shorthand form is **idiomatic Go** and more concise. However, if maintainers prefer explicit types for consistency with other tests in this codebase, I'm happy to change it.

asreenath:
## Response to Suppressed Comment: Invalid Values Type Handling

**Excellent catch!** This is a valid concern.

**Current behavior:**
When `values` is not an array (e.g., `"values":"prod"` or `"values":123`), the code silently ignores it and creates a key-only filter, which matches resources with ANY value for that key.

**Why this is problematic:**
- User writes `{"key":"Environment","values":"prod"}` (typo/mistake)
- Expects: resources with Environment=prod
- Gets: resources with Environment=ANY_VALUE (overly broad)
- No error/warning to indicate the mistake

**Proposed fix:**
Return an error when `values` exists but is not an array type. This provides clear feedback instead of silent fallback.

```go
if valuesRaw, exists := filter["values"]; exists {
    valuesArray, ok := valuesRaw.([]interface{})
    if !ok {
        return nil, fmt.Errorf("'values' must be an array of strings, got %T", valuesRaw)
    }
    // ... rest of values processing
}
```

**Impact:**
- ✅ Fail-fast on invalid input (better UX)
- ✅ Prevents unexpected overly-broad queries
- ✅ Consistent with existing error handling (invalid JSON returns error)

I'll add this validation + test case. Would you like me to push this as a new commit?

asreenath:
## Implemented Fix for Invalid Values Type Validation

I've addressed the bot's concern about silent fallback on invalid `values` types.

**Changes in commit 3ccf4a04:**

### 1. Added `fmt` import
Required for error formatting with `fmt.Errorf`.

### 2. Added type validation with fail-fast error
```go
if valuesRaw, exists := filter["values"]; exists {
    valuesArray, ok := valuesRaw.([]interface{})
    if !ok {
        return nil, fmt.Errorf("tag_filter: 'values' must be an array of strings, got %T", valuesRaw)
    }
    // ... continue processing
}
```

**Before:** `{"key":"Env","values":"prod"}` silently ignored → key-only match (too broad)  
**After:** Returns clear error: `tag_filter: 'values' must be an array of strings, got string`

### 3. Added test cases
- Test for string type: `{"key":"Environment","values":"prod"}` → expect error ✅
- Test for number type: `{"key":"Environment","values":123}` → expect error ✅

### Results
✅ All 13 unit tests pass (11 original + 2 new)  
✅ Compilation succeeds with no errors  
✅ Fail-fast prevents user typos from causing unexpected overly-broad queries

**Impact:**
- Better user experience: immediate clear error instead of silent wrong results
- Consistent with existing error handling (invalid JSON also returns error)
- Prevents production issues from typos like `"values":"prod"` instead of `"values":["prod"]`

aws/table_aws_tagging_resource.go
```diff
@@ -66,6 +72,12 @@ func tableAwsTaggingResource(_ context.Context) *plugin.Table {
 				Type:        proto.ColumnType_JSON,
 				Transform:   transform.FromField("Tags"),
 			},
+			{
+				Name:        "tag_filter",
+				Description: "A list of TagFilters used to filter resources by tags. Specify a JSON array of objects with 'key' and optional 'values' fields, e.g., [{\"key\":\"Environment\",\"values\":[\"prod\",\"dev\"]}].",
```
cbruno10, line 77: I used this filter in queries and it works well: a single key, key plus values, and an `in` list of filters all matched the client-side counts exactly.

Can you please also add examples to `docs/tables/aws_tagging_resource.md`? Plain string literals are the way to go, that's what every other JSON qual in the repo does and it reads the same in both dialects (no `::jsonb` cast needed, the column is already jsonb). Four shapes would cover it:

```sql
-- resources that have the tag, any value
where tag_filter = '[{"key": "Environment"}]'

-- one of several values (ORed)
where tag_filter = '[{"key": "Environment", "values": ["prod", "staging"]}]'

-- two tags that must both be present (ANDed)
where tag_filter = '[{"key": "Environment", "values": ["prod"]}, {"key": "Owner"}]'

-- alternative filters, unioned
where tag_filter in ('[{"key": "Team", "values": ["platform"]}]', '[{"key": "Team", "values": ["data"]}]')
```

aws/table_aws_tagging_resource.go
```diff
@@ -173,6 +195,58 @@ func getTaggingResource(ctx context.Context, d *plugin.QueryData, _ *plugin.Hydr
...
+			if val != "" && q.Operator == "=" {
+				// Parse JSON: [{"key":"environment","values":["prod","dev"]},{"key":"team"}]
+				var filterInput []map[string]interface{}
+				if err := json.Unmarshal([]byte(val), &filterInput); err != nil {
+					return nil, err
+				}
+
+				for _, filter := range filterInput {
+					keyStr, keyOk := filter["key"].(string)
+					if !keyOk || keyStr == "" {
```
cbruno10: This is the one I'd change. Anything without a usable `key` is dropped silently, so `[]`, `[{"values":["x"]}]` and `[{"Key":"Name"}]` (capital K) all returned every resource in the account when I tried them, which is the worst outcome for a filter typo. Could we return an error here instead of `continue`, and also when the array parses but produces no filters?

aws/table_aws_tagging_resource.go
```diff
@@ -173,6 +195,58 @@ func getTaggingResource(ctx context.Context, d *plugin.QueryData, _ *plugin.Hydr
...
+
+				// Values are optional
+				if valuesRaw, exists := filter["values"]; exists {
+					valuesArray, ok := valuesRaw.([]interface{})
+					if !ok {
+						return nil, fmt.Errorf("tag_filter: 'values' must be an array of strings, got %T", valuesRaw)
+					}
+					var values []string
+					for _, v := range valuesArray {
+						if vStr, vOk := v.(string); vOk && vStr != "" {
```
cbruno10: Now that a non-array `values` errors, non-string entries should probably error too rather than being skipped. `["prod",1,true]` quietly became `["prod"]` in my test.

One way to get all of this for free is to unmarshal into `[]struct{ Key string; Values []string }`: the JSON decoder then rejects wrong types with a clear message and we lose the manual type switching.

aws/table_aws_tagging_resource.go
```diff
@@ -173,6 +195,58 @@ func getTaggingResource(ctx context.Context, d *plugin.QueryData, _ *plugin.Hydr
...
+func buildTagFilter(d *plugin.QueryData) ([]types.TagFilter, error) {
+	var tagFilters []types.TagFilter
+
+	if d.Quals["tag_filter"] != nil {
+		for _, q := range d.Quals["tag_filter"].Quals {
+			val := q.Value.GetJsonbValue()
+			if val != "" && q.Operator == "=" {
+				// Parse JSON: [{"key":"environment","values":["prod","dev"]},{"key":"team"}]
+				var filterInput []map[string]interface{}
+				if err := json.Unmarshal([]byte(val), &filterInput); err != nil {
```
cbruno10: The raw decoder error is a bit opaque to a user (`json: cannot unmarshal object into Go value of type []map[string]interface {}`). Wrapping it with `fmt.Errorf("tag_filter: %w", err)` would say which qual is at fault.

aws/table_aws_tagging_resource.go
```diff
@@ -173,6 +195,58 @@ func getTaggingResource(ctx context.Context, d *plugin.QueryData, _ *plugin.Hydr
...
+					keyStr, keyOk := filter["key"].(string)
+					if !keyOk || keyStr == "" {
+						continue
+					}
+
+					tagFilter := types.TagFilter{
+						Key: aws.String(keyStr),
+					}
+
+				// Values are optional
```
cbruno10: This block is one tab shallow, and the test file has the same at lines 72 to 89. Lint doesn't run gofmt here so CI won't catch it; can you run `gofmt -w` on both files?

aws/table_aws_tagging_resource.go
```diff
@@ -173,6 +195,58 @@ func getTaggingResource(ctx context.Context, d *plugin.QueryData, _ *plugin.Hydr
...
+				}
+
+					tagFilters = append(tagFilters, tagFilter)
+				}
+			}
+		}
+	}
+
+	return tagFilters, nil
+}
```
cbruno10:
```suggestion
// buildTagFilter constructs TagFilter objects from the tag_filter qual
func buildTagFilter(d *plugin.QueryData) ([]types.TagFilter, error) {
	if d.Quals["tag_filter"] == nil {
		return nil, nil
	}

	var tagFilters []types.TagFilter
	for _, q := range d.Quals["tag_filter"].Quals {
		val := q.Value.GetJsonbValue()
		if val == "" || q.Operator != "=" {
			continue
		}

		// Decode straight into the SDK type, e.g. [{"key":"environment","values":["prod","dev"]},{"key":"team"}]
		var filters []types.TagFilter
		if err := json.Unmarshal([]byte(val), &filters); err != nil {
			return nil, fmt.Errorf("tag_filter: %w", err)
		}
		if len(filters) == 0 {
			return nil, fmt.Errorf("tag_filter: expected a non-empty array of {\"key\", \"values\"} objects")
		}
		for i, f := range filters {
			if f.Key == nil || *f.Key == "" {
				return nil, fmt.Errorf("tag_filter: entry %d is missing \"key\"", i)
			}
			if len(f.Values) == 0 {
				filters[i].Values = nil
			}
		}

		tagFilters = append(tagFilters, filters...)
	}

	return tagFilters, nil
}
```

Suggest decoding straight into `[]types.TagFilter` the way `aws_cloudwatch_metric` does with `[]types.DimensionFilter`. The decoder then does the type checking for us: `"values": ["prod", 1]` fails with a clear message instead of quietly becoming `["prod"]`, both `key` and `Key` work since field matching is case-insensitive, and the manual type switching goes away.

The one thing it doesn't catch is a missing key, and that's the case I'd most like to error on. When I tried `[]`, `[{"values":["x"]}]` and `[{"Key":"Name"}]` on the branch, all three silently returned every resource in the account, which is the worst outcome for a filter typo. The two explicit checks above cover that.

I compiled and vetted this version against the branch. Your unit tests will need updating for the stricter behaviour (the missing-key cases now error, and an empty string in `values` is kept rather than dropped, which matches how the API treats empty tag values). The test file also needs a `gofmt -w`, lines 72 to 89 are a tab shallow and lint here doesn't check formatting.

aws/table_aws_tagging_resource.go
```diff
@@ -27,6 +30,9 @@ func tableAwsTaggingResource(_ context.Context) *plugin.Table {
 		List: &plugin.ListConfig{
 			Hydrate: listTaggingResources,
 			Tags:    map[string]string{"service": "tag", "action": "GetResources"},
+			KeyColumns: plugin.KeyColumnSlice{
+				{Name: "tag_filter", Operators: []string{"="}, Require: plugin.Optional, CacheMatch: query_cache.CacheMatchExact},
```
cbruno10:
```suggestion
				{Name: "tag_filters", Operators: []string{"="}, Require: plugin.Optional, CacheMatch: query_cache.CacheMatchExact},
```

Small naming thought while it's still cheap, the SDK field is `TagFilters` and the CLI flag is `--tag-filters`, and the qual takes an array of them, so `tag_filters` is what someone coming from the CLI will guess. If we update it here, we'll also need updates in `FromQual`, and in `buildTagFilter`.

cbruno10 (changes requested):
@asreenath Thanks for the PR! I've left a few suggestions, can you please have a look? Thanks!

asreenath:
## Response to @cbruno10's Review

Thanks for the detailed review and for verifying the filter behavior against client-side counts - that's really valuable confirmation.

All three suggestions implemented in commit 3b6536b0:

**1. Renamed `tag_filter` → `tag_filters`**
Matches the SDK field name (`TagFilters`) and CLI convention (`--tag-filters`). Updated the column definition, `KeyColumns`, `FromQual`, `buildTagFilter`, unit tests, and the integration test query.

**2. Rewrote `buildTagFilter` to decode directly into `[]types.TagFilter`**
Adopted your suggested implementation exactly - decoding straight into the SDK type instead of manual `map[string]interface{}` parsing. This:
- Lets the JSON decoder enforce value types (e.g. `"values":["prod",1]` now fails clearly instead of silently becoming `["prod"]`)
- Adds the two explicit checks you flagged: empty/missing `key` and empty filter array both now return errors instead of silently matching every resource
- Removes the manual type-switching code entirely

**3. Added docs examples**
Added the four example shapes you listed to `docs/tables/aws_tagging_resource.md` (single key, OR values, AND across keys, unioned via `IN`), using plain string literals as you suggested.

**Test updates:**
- Updated existing cases for the stricter behavior (missing/empty key now expects an error, empty string in `values` is now kept rather than dropped)
- Added cases for mixed-type arrays and empty filter arrays
- Ran `gofmt -w` on both files
- All 15 unit tests pass, `go build ./aws` is clean

One thing to flag: since this renames the key column, if this PR has already been shared with anyone testing against `tag_filter`, they'll need to switch to `tag_filters`. Given the PR hasn't merged yet, I think now is the right time to make this change. Let me know if you'd like anything else adjusted.

asreenath:
Missed one spot on the `::jsonb` note: the integration test query (`aws-test/tests/aws_tagging_resource/test-tag-filter-query.sql`) still had the cast. Removed it in commit 1dcc39a1 - now uses a plain string literal, matching the docs examples.

cbruno10 (approved):

